PR #195, opened 2015-01-04
ghc 7.10.1 RC1 requires FlexibleContexts

markwright:
https://ghc.haskell.org/trac/ghc/wiki/Migration/7.10

"Compiler changes
Inferred type-signatures now may require to enable FlexibleContexts..."

Compiles with ghc 7.10.1 RC1 and 7.8.4.


snoyberg:
Thanks!

PR #201, opened 2015-02-22
conduit-extra: Add Data.Conduit.ByteString.Builder

manny-fp:


manny-fp:
I started this before blaze-builder-0.4 was released, to make it easy to use both types of `Builder`.  But now that blaze-builder uses the same Builder datatype as bytestring and streaming-commons is updated to support it, this isn't really necessary anymore, since Data.Conduit.Blaze will work with bytestring's `Builder` when using blaze-builder-0.4 anyway.  That means the only value here is the new module name to reflect a `blaze-builder`-less future, and the dubious ability for projects still using blaze-builder-0.3 to also use bytestring's `Builder`.

PR #203, opened 2015-03-10
Support chain certificates in network-conduit-tls

ygale:
This compiles, but it is untested. I believe it ought to work though.

### DIFF
--- a/network-conduit-tls/Data/Conduit/Network/TLS.hs
+++ b/network-conduit-tls/Data/Conduit/Network/TLS.hs
@@ -9,6 +9,8 @@ module Data.Conduit.Network.TLS
       TLSConfig
     , tlsConfigBS
     , tlsConfig
+    , tlsConfigChainBS
+    , tlsConfigChain
     , tlsHost
     , tlsPort
 --    , tlsCertificate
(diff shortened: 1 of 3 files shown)


network-conduit-tls/Data/Conduit/Network/TLS.hs
```diff
@@ -58,19 +60,24 @@ import Control.Monad.Trans.Control
 import Data.Default
 
 
-makeCertDataPath :: FilePath -> FilePath -> TlsCertData
-makeCertDataPath certPath keyPath = TlsCertData (readFile certPath) (readFile keyPath)
-
-makeCertDataBS :: S.ByteString -> S.ByteString -> TlsCertData
-makeCertDataBS certBS keyBS = TlsCertData (return certBS) (return keyBS)
+makeCertDataPath :: FilePath -> [FilePath] -> FilePath -> TlsCertData
```
snoyberg: This makes it a breaking change, which I'd rather avoid.

ygale:
Those functions are not exported, so I didn't view that as breaking change. But it is of course possible to create separate versions of them to use in the new smart constructors. Is there a reason why it's important to do that?

snoyberg:
Sorry, my mistake, I thought they were exported. I'll continue my review,
thanks.

snoyberg:
Thanks!

PR #207, opened 2015-03-29
Add split function to Data.Text.Conduit (closes #205)

### LINKED ISSUE #205: split function for Data.Conduit.Text
Hi, I needed a split function for some log parsing I was doing with Conduit, would you be interested in me adding this to `Data.Conduit.Text`? If so, I can add tests and remove the Safe dependency that my code is using, then submit a PR.

This is the code I'm currently using; it's modeled closely after `lines` and I believe `lines` could be implemented in terms of `split`.

```haskell
split :: Monad m => Text -> Conduit Text m Text
split splitText = awaitText T.empty
    where
      awaitText buf = await >>= maybe (finish buf) (process buf)

      finish buf = unless (T.null buf) (yield buf)

      process buf text = yieldSplits $ buf `T.append` text

      yieldSplits text = do
        let splits = T.splitOn splitText text
            lastSplit = lastDef T.empty splits
        mapM_ yield (initSafe splits)
        awaitText lastSplit
```

MaxGabriel:
Don't merge; this is missing changelog updates and documentation. I just thought these tests could address @snoyberg's [comment](https://github.com/snoyberg/conduit/issues/205#issuecomment-87275036) about not splitting correctly when the input is chunked. This test in particular:

```haskell
it "handles separators on a chunk boundary" $ do
    (CL.sourceList ["aX","Xb"] C.$= CT.split "XX" C.$$ CL.consume) ==
        [["a","b"]]
```

### DIFF
--- a/conduit-extra/test/Data/Conduit/TextSpec.hs
+++ b/conduit-extra/test/Data/Conduit/TextSpec.hs
@@ -7,6 +7,7 @@ import qualified Data.Conduit.Lift as C
 import qualified Data.Conduit.List as CL
 import Test.Hspec
 import Test.Hspec.QuickCheck
+import Test.QuickCheck.Modifiers
 import Data.Monoid
 import Control.Monad.ST
 import qualified Data.Text as T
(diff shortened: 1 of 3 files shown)


conduit-extra/test/Data/Conduit/TextSpec.hs
```diff
@@ -187,6 +188,44 @@ spec = describe "Data.Conduit.Text" $ do
...
+                [["",""]] -- NB: Data.Text.splitOn would have given 3 empty strings.
+                          -- The 2 empty strings behavior is how `Data.Conduit.Text.lines` / `Data.String.lines` works, though
+        prop "works the same no matter how things are chunked" $ \s1 s2 int -> do
+            let t1 = T.pack s1
+                t2 = T.pack (getNonEmpty s2)
+                positiveInt = getPositive int
+                splitUp = T.chunksOf positiveInt t1
+            l1 <- (CL.sourceList [t1] C.$= CT.split t2 C.$$ CL.consume)
+            l2 <- (CL.sourceList splitUp C.$= CT.split t2 C.$$ CL.consume)
+            l1 `shouldBe` l2
```
MaxGabriel: I've pretty much never used QuickCheck before, so I'm not sure how valuable this test is. I can replace it with manual tests checking the same thing, if necessary.

snoyberg:
OK, I see it now, you're correct. The problematic thing is that this will take up unbounded memory unfortunately, which ideally would not be necessary. An algorithm like those available in http://www.stackage.org/package/stringsearch would be necessary to avoid that.

MaxGabriel:
@snoyberg I don't fully follow, could you clarify? It seems like taking up unbounded memory is inherent to this type of function since the separator may never come, and the text read so far has to be kept around in case it does. `lines` has the same issue in the case where [there's no newline separator](https://github.com/snoyberg/conduit/blob/c59206d52cf7c7d8add1bb44097d8e47fe965a1b/conduit-extra/Data/Conduit/Text.hs#L101). 

It seems like the algorithms in stringsearch could help with running time performance but not necessarily memory usage?

On the other hand, I am getting really high memory usage when running this function on a 50 MB file that doesn't have the specified separator in it.

snoyberg:
Imagine you're looking to break on the string "abc". If I get a chunk "qwertya", I know that "qwerty" will be before the chunk boundary, but "a" may be the beginning of it. I can therefore break it off, send it downstream, and continue. If the next chunk is "bc...", then I've found a boundary. If it starts with anything else, I know that it's not a boundary.

You're completely correct about `lines` having unbounded memory usage, which is why it should only be used in data that you have control over (or with some extra method to ensure memory protection). Instead, a function like [line](https://www.stackage.org/haddock/lts-1.15/conduit-combinators-0.3.0.6/Data-Conduit-Combinators.html#v:line) is probably a better interface for a function like this, since it allows for completely deterministic memory usage.

gregwebs:
I will attempt to summarize:

@snoyberg is pointing out that `splitOn` is not streaming: it builds up a list of results
@MaxGabriel is pointing out that in terms of memory consumption that should only worst case double the memory consumption because the original string is still in memory

It seems to me that

  * to avoid unbounded memory usage this would need to ask for a text of bounded length.
  * this function should avoid the doubling of memory and send results back immediately.

MaxGabriel:
@gregwebs I'm not positive that's what @snoyberg was saying, but I don't think that `Data.Text.splitOn` doubles memory usage. [Its implementation](https://github.com/bos/text/blob/62e0bb7b0912572cb1fb4e760a6143dffe99dc56/Data/Text.hs#L1336-L1337) takes in a `Text` value (backed by an `Array`) to be split up. Then, instead of creating new arrays for the split-up substrings, it uses the original array + different indices for the beginning and end of the text.

I think that's that this line [from the Data.Text haddocks](http://hackage.haskell.org/package/text-1.2.0.4/docs/Data-Text.html#g:18) is getting at:

> Splitting functions in this library do not perform character-wise copies to create substrings; they just construct new Texts that are slices of the original.

snoyberg:
To clarify my thoughts: the currently proposed API necessitates collecting multiple chunks into memory until it finds a boundary. This defeats the streaming nature of conduit. Instead, I'd prefer an API that provides individual streams for each piece of text between chunks. This is very similar to the difference between the `lines` and `line` functions in conduit-combinators.

gregwebs:
@MaxGabriel thanks for pointing to the docs. I think the worst case could still be greater than 2x memory usage though: if every other character is split that creates a lot of new Text objects.

PR #208, opened 2015-04-20
Added 'runGeneralTCPServerTLS' function

anton-dessiatov:
This function is almost exactly 'runGeneralTCPServer' from Data.Conduit.Network (conduit-extra) and serves the same purpose - generalizes runTCPServerTLS to any monad, which is an instance of MonadBaseControl IO.


snoyberg:
Thanks!

PR #209, opened 2015-04-21
Make Binary.lines O(n) instead of O(n^2)

ndmitchell:
Given the benchmark:

    import Data.Conduit.Binary as C
    import Data.Conduit as C
    import Data.Conduit.List as C
    import Data.Void
    import qualified Data.ByteString.Char8 as BS
    import Control.Monad
    import Control.Monad.IO.Class
    import System.Time.Extra
    import Numeric.Extra

    main = do
        let s = BS.singleton 'x'
        forM_ [0..1000] $ \i -> do
            (t, res) <- duration $ runConduit $ replicateM_ (i*10000) (yield s) =$= C.lines =$= C.map BS.length =$= C.fold (+) 0
            print (showDuration $ t / intToDouble i, res)

In other words, generate a lot of short bytestrings, call lines (which basically concats it, since there are no newlines), count the total number of lines and print the time per 10K characters, I see:

* Before this patch: at 10K it takes 0.01s, at 300K it takes 0.13s per 10K elements.
* After this patch: it takes 0.00s per 10K elements at all numbers.

The cause of the poor performance before was an O(n^2) term. Someone did the "append as a function trick" - which kinda works for lists, but doesn't work at all for bytestrings. I did the simpler (and in my experience, more performant) list that you reverse construction, which actually works as you then call `S.concat` once.

### DIFF
--- a/conduit-extra/Data/Conduit/Binary.hs
+++ b/conduit-extra/Data/Conduit/Binary.hs
@@ -345,20 +345,18 @@ drop n0 = go n0
 -- Since 0.3.0
 lines :: Monad m => Conduit S.ByteString m S.ByteString
 lines =
-    loop id
+    loop []
   where
     loop front = await >>= maybe (finish front) (go front)
 
     finish front =
-        let final = front S.empty
+        let final = S.concat $ reverse front
          in unless (S.null final) (yield final)
 
     go sofar more =
         case S.uncons second of
-            Just (_, second') -> yield (sofar first) >> go id second'
-            Nothing ->
-                let rest = sofar more
-                 in loop $ S.append rest
+            Just (_, second') -> yield (S.concat $ reverse $ first:sofar) >> go [] second'
+            Nothing -> loop $ more:sofar
       where
         (first, second) = S.breakByte 10 more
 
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks. I haven't run it myself, but I'll trust your benchmark numbers. This is the first I've heard of a O(n^2) generated by using difference lists, is this a well known phenomenon?

ndmitchell:
Difference lists tend to suck performance wise due to a high constant factor (thunks do not have zero cost). In this case the difference list wasn't the issue, but the lots of appends. This patch transforms:

    (a `S.append` b) `S.append` c

Into:

    S.concat [a,b,c]

Which does one allocation for the entire size, instead of n allocations of increasingly long prefixes of the list (which in the limit take n), making n^2. It doesn't matter whether you run the appends left or right associated, they're always n^2 as `S.append a b` takes _O(a+b)_. In contrast, for lists/String, `a ++ b` takes _O(a)_ so if you associate the right way you get _O(n)_.

snoyberg:
Oh, of course, I just misread what you wrote originally. Thanks for the clarification.

PR #213, opened 2015-05-12
Adding buffer size to sourceHandleRange

nhibberd:
We're reading from disk and uploading parts of files using http-conduit and we've observed that the buffer size has significant impact on performance. On certain machines going from a 4k buffer to a 1mb buffer results in a 4x speed up.

Not 100% on the signature that you would prefer, this lets us achieve what we need. If you would like it in a slightly different form just let me know.

### DIFF
--- a/conduit-extra/Data/Conduit/Binary.hs
+++ b/conduit-extra/Data/Conduit/Binary.hs
@@ -17,6 +17,7 @@ module Data.Conduit.Binary
     , sourceIOHandle
     , sourceFileRange
     , sourceHandleRange
+    , sourceHandleRangeWithBuffer
       -- ** Sinks
     , sinkFile
     , sinkHandle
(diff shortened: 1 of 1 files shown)


conduit-extra/Data/Conduit/Binary.hs
```diff
@@ -181,7 +182,20 @@ sourceHandleRange :: MonadIO m
                   -> Maybe Integer -- ^ Offset
                   -> Maybe Integer -- ^ Maximum count
                   -> Producer m S.ByteString
-sourceHandleRange handle offset count = do
+sourceHandleRange handle offset count =
+  sourceHandleRangeWithBuffer handle offset count 4096
```
snoyberg: Ouch, having the buffer size so low in the first place was a big mistake. I'll fix that also.

snoyberg:
LGTM :+1:

PR #214, opened 2015-05-16
Add pass-through instances for Alternative and MonadPlus

k0ral:


resourcet/Control/Monad/Trans/Resource/Internal.hs
```diff
@@ -235,6 +236,14 @@ instance Applicative m => Applicative (ResourceT m) where
     ResourceT mf <*> ResourceT ma = ResourceT $ \r ->
         mf r <*> ma r
 
+instance Alternative m => Alternative (ResourceT m) where
+    empty = ResourceT $ \r -> empty
```
snoyberg: Can this be changed to `\_ -> empty` to avoid a warning?

resourcet/Control/Monad/Trans/Resource/Internal.hs
```diff
@@ -235,6 +236,14 @@ instance Applicative m => Applicative (ResourceT m) where
...
+    (ResourceT mf) <|> (ResourceT ma) = ResourceT $ \r -> mf r <|> ma r
+
+instance MonadPlus m => MonadPlus (ResourceT m) where
+    mzero = ResourceT $ \r -> mzero
```
snoyberg: Same here.

k0ral:
Fixed the warnings.

snoyberg:
Thanks!

PR #217, opened 2015-05-26
Added detectUtf function to module Data.Conduit.Text

### LINKED ISSUE #216: detectUtf function for Data.Conduit.Text
I'm consuming a data feed that has a UTF-8 BOM (0xEF, 0xBB, 0xBF), and I want to remove that if it exists, because otherwise my sink-parser doesn't work correctly (`decode utf8` doesn't remove BOM). Right now I'm using this function:

```haskell
import qualified Data.ByteString              as BS
import qualified Data.Conduit.Text            as CT

detectUtf :: MonadThrow m => Conduit BS.ByteString m T.Text
detectUtf = do
  mb <- await
  case mb of
    Just b -> do
      let (toDrop, codec) =
            case BS.unpack b of
              0x00 : 0x00 : 0xFE : 0xFF : _ -> (4, CT.utf32_be)
              0xFF : 0xFE : 0x00 : 0x00 : _ -> (4, CT.utf32_le)
              0xFE : 0xFF : _               -> (2, CT.utf16_be)
              0xFF : 0xFE : _               -> (2, CT.utf16_le)
              0xEF : 0xBB : 0xBF : _        -> (3, CT.utf8)
              _                             -> (0, CT.utf8)
      leftover . BS.drop toDrop $ b
      CT.decode codec
    Nothing -> return ()
```

I'm not sure whether I need to call the await function as long as the total length of consumed bytestrings < 4, but it works in my case.
It would be nice to have correct implementation of this function in conduit-extra package.

dimchansky:
this commit resolves #216

### DIFF
--- a/conduit-extra/Data/Conduit/Text.hs
+++ b/conduit-extra/Data/Conduit/Text.hs
@@ -35,6 +35,7 @@ module Data.Conduit.Text
     , Data.Conduit.Text.decodeUtf8
     , decodeUtf8Lenient
     , encodeUtf8
+    , detectUtf
     ) where
 
 import qualified Prelude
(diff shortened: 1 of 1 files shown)


snoyberg:
Thanks!

PR #220, opened 2015-07-30
Safe/Trustworthy for resourcet

rimmington:
This makes it possible to use resourcet in Safe Haskell. I've removed a bunch of imports that appear (and are reported as) redundant, but suspect I'm missing something.

Tested with LTS Haskell 2.20 (GHC 7.8.4), but no other versions.


rimmington:
Unrelated: another compiler warning notes that `Control.Monad.Trans.Resource.finally` is declared but unused, which appears to be the case.

snoyberg:
Thanks!

PR #229, opened 2015-10-07
 Bring README type variable names into line with sources.

haskellGardener:
Bring README type variable names into line with sources.


snoyberg:
Thanks!

PR #231, opened 2015-10-16
Fix inconsistent void dependency

peti:
`void` must be included only when compiling with GHC versions prior to 7.9.x. The library had that restriction already, but it was missing in the spec of the test suite.


snoyberg:
Thanks!

PR #234, opened 2015-11-01
Update README.md

ra1u:


snoyberg:
Is this necessary to get the example to compile?

ra1u:
that was for me -> error
pg.hs:12:5: Not in scope: ‘runResourceT’

my version is 
conduit: 1.2.5.1
conduit-combinators: 1.0.3
but i havent checked other examples

snoyberg:
It would be better to import from Control.Monad.Trans.Resource

ra1u:
I can check other examples and update PR

snoyberg:
Looks like this change has already made it to master.

PR #235, opened 2015-11-02
Add mergeSource (closes #233)

### LINKED ISSUE #233: Helper function that injects a Source into a Conduit
Hi there,

I want to inject a `Source` into a `Conduit`. I think it may look like the opposite side of `passthroughSink`.
That should be something like:

```Haskell
-- | Inject a @Source@ into a @Conduit@.
-- The new conduit will stop processing once either source or upstream have been exhausted.
injectSource :: Monad m => Source m i -> Conduit a m (i, a)
injectSource = injectResumableSource . newResumableSource

injectResumableSource :: Monad m => ResumableSource m i -> Conduit a m (i, a)
injectResumableSource src = await >>= maybe (lift $ closeResumableSource src) go
  where
    go a = do
      (src1, mi) <- lift $ src $$++ await
      case mi of
        Nothing -> lift $ closeResumableSource src1
        Just i  -> yield (i, a) >> injectResumableSource src1
```

And an example usage:

```Haskell
import Data.Conduit
import Control.Monad.Trans
import Text.Printf
import qualified Data.ByteString.Char8 as C
import qualified Data.Conduit.Binary as CB
import qualified Data.Conduit.List as CL

withLineNumber :: Monad m => Conduit i m (Integer, i)
withLineNumber = injectSource (CL.sourceList [1..])

main :: IO ()
main = runResourceT
  $  CB.sourceFile filename
  $= CB.lines
  $= CL.map C.unpack
  $= withLineNumber
  $$ CL.mapM_ (\(i, l) -> liftIO $ printf "%06d: %s\n" i l)
```

Would you like to add something like this?

Thanks.

tim-semba:
See further comments in issue #233 

Please let me know when you find more better names.
I think the type signature looks like typical `zip`, but `zipSource` confuses us with `zipSources`.

Thanks.

### DIFF
--- a/conduit/Data/Conduit.hs
+++ b/conduit/Data/Conduit.hs
@@ -53,6 +53,7 @@ module Data.Conduit
     , mapOutputMaybe
     , mapInput
     , passthroughSink
+    , sourceToList
 
       -- * Connect-and-resume
     , ResumableSource
--- a/conduit/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/Data/Conduit/Internal/Conduit.hs
@@ -673,7 +673,7 @@ mergeSource
   :: Monad m
   => Source m i
   -> Conduit a m (i, a)
-mergeSource src = loop $ newResumableSource src
+mergeSource = loop . newResumableSource
   where
     loop :: Monad m => ResumableSource m i -> Conduit a m (i, a)
     loop src0 = await >>= maybe (lift $ closeResumableSource src0) go
--- a/conduit/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/Data/Conduit/Internal/Conduit.hs
@@ -75,6 +75,7 @@ module Data.Conduit.Internal.Conduit
     , zipSourcesApp
     , zipConduitApp
     , passthroughSink
+    , sourceToList
     , fuseBoth
     , fuseBothMaybe
     , fuseUpstream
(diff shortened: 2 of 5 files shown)


snoyberg:
Two concerns:

* I'm not too happy with the name, but unfortunately I can't think of something better.
* I think something like `addCleanup` will be necessary to ensure that the finalizers are run correctly

tim-semba:
> I'm not too happy with the name, but unfortunately I can't think of something better.

Some ideas:

* `flowintoConduit` (contrast with `passthroughSink`)
* `mergeintoConduit`
* `joinSource` (variant of `mergeSource`)


> I think something like addCleanup will be necessary to ensure that the finalizers are run correctly

I tried to fix, but I'm afraid I don't really understand what it is.
I hope the `Source`s can handle the finalizers correctly by themselves, but I don't know how to do so.

Thanks.

snoyberg:
I don't think the extra argument providing a finalizer is necessary (and sorry if I implied that it was previously). Your new code does handle the finalizer case correctly. I've just pushed two commits to the tim-semba-master branch including a test case for the behavior I was worried about. Do you want to update your branch one more time with the simplified signature and the test case?

tim-semba:
Thank you very much.
I fixed the tests and pushed it.

Thanks.

snoyberg:
Thanks!

PR #241, opened 2015-12-25
Import runResourceT to example

technocrat:
to example in README.md needed to bring runResourceT into scope for
runResourceT $ CB.sourceFile "input.txt" $$ CB.sinkFile "output.txt"


snoyberg:
Thanks

PR #247, opened 2016-02-08
Fix conduit and resourcet on GHC 8.0

RyanGlScott:
This requires two small changes:

* The `ImpredicativeTypes` extension pretty much can't be used on GHC 8.0 due to a [type inference regression](https://ghc.haskell.org/trac/ghc/ticket/11319). `ImpredicativeTypes` has been pretty much unsupported for a while though, so removing it is probably for the best.
* Without `ImpredicativeTypes`, the definition of `resourceMask` had to be expanded.


snoyberg:
Thanks!

PR #251, opened 2016-02-13
Proposed `fuseEither` combinator

DaveCTurner:
As discussed on [Haskell Cafe](https://mail.haskell.org/pipermail/haskell-cafe/2016-February/123132.html) I'd quite like to be able to fuse two conduits together and return the result of whichever one exits first.

The attached PR represents a bit of a lucky guess, in the sense that it compiles and passes a few sanity-check tests, but I don't know whether it really makes sense or whether it does something horrible in more advanced cases. It's also basically a copy-paste job from the definition of `(=$=)` and currently doesn't have documentation, both of which I'm happy to fix with a bit of guidance.

### DIFF
--- a/conduit/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/Data/Conduit/Internal/Conduit.hs
@@ -829,7 +829,7 @@ fuseEither (ConduitM left0) (ConduitM right0) = ConduitM $ \rest ->
             case left of
                 HaveOutput left' final' o -> goRight final' left' (rp o)
                 NeedInput left' lc        -> NeedInput (recurse . left') (recurse . lc)
-                Done r1                   -> PipeM (final >> return (rest (Left r1)))
+                Done r1                   -> PipeM (return (rest (Left r1)))
                 PipeM mp                  -> PipeM (liftM recurse mp)
                 Leftover left' i          -> Leftover (recurse left') i
           where
--- a/conduit/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/Data/Conduit/Internal/Conduit.hs
@@ -80,6 +80,7 @@ module Data.Conduit.Internal.Conduit
     , fuseBoth
     , fuseBothMaybe
     , fuseUpstream
+    , fuseEither
     , sequenceSources
     , sequenceSinks
     , sequenceConduits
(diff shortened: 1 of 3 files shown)


conduit/Data/Conduit/Internal/Conduit.hs
```diff
@@ -812,6 +813,29 @@ ConduitM left0 =$= ConduitM right0 = ConduitM $ \rest ->
...
+                PipeM mp          -> PipeM (liftM recurse mp)
+                Leftover right' i -> goRight final (HaveOutput left final i) right'
+          where
+            recurse = goRight final left
+
+        goLeft rp rc final left =
+            case left of
+                HaveOutput left' final' o -> goRight final' left' (rp o)
+                NeedInput left' lc        -> NeedInput (recurse . left') (recurse . lc)
+                Done r1                   -> PipeM (final >> return (rest (Left r1)))
```
snoyberg: You don't want to call `final` here; once upstream terminates, all of its finalizers must already be called.

conduit/Data/Conduit/Internal/Conduit.hs
```diff
@@ -812,6 +813,29 @@ ConduitM left0 =$= ConduitM right0 = ConduitM $ \rest ->
   where
 {-# INLINE [1] (=$=) #-}
 
+fuseEither :: Monad m => ConduitM a b m u -> ConduitM b c m d -> ConduitM a c m (Either u d)
```
snoyberg: There should be a Haddock comment on this function explaining its purpose and how it's different from `=$=`.

DaveCTurner:
Thanks Michael. I'll get to the docs later. You're absolutely right about not wanting the call to `final` there - I thought that would run the downstream finalisers but apparently not. However, in testing this I've discovered that in this implementation if the upstream exits first then the downstream's finalisers aren't run at all. I've added a couple of test cases, one of them failing, showing what I mean. Is this fixable?

snoyberg:
You're right, this is not handled. An invariant of how conduit works is that downstream always has an opportunity to clean up itself, and therefore it does not have finalizers sent upstream. What you're trying to do is essentially recapture a capability of the `Pipe` layer of abstraction inside `ConduitM`, which can't reasonably be done. Is there a reason you're not just dropping down to using `Pipe` for this specific case?

DaveCTurner:
> Is there a reason you're not just dropping down to using Pipe for this specific case?

Mainly the belief that it *could* be done! Never mind, I'll do it another way :)

PR #256, opened 2016-02-27
Improve documentation of bracketP

harendra-kumar:
I was reading it for the first time and found it a bit unclear.


snoyberg:
Thanks!

Issue #250, opened 2016-02-12
Capture both stdout and stderr with Data.Conduit.Process?


joehillen:
I keep finding myself is situations were I need to capture both stdout and stderr in parallel.

I've implemented it 2 different ways. Both "work", but both have problems.

First is this way:

```haskell
sourceProcessWithConsumers :: CreateProcess
                           -> Consumer ByteString IO () -- stdout
                           -> Consumer ByteString IO () -- stderr
                           -> IO ExitCode
sourceProcessWithConsumers cp consumerStdout consumerStderr = do
    ( ClosedStream, (sourceStdout, closeStdout)
                  , (sourceStderr, closeStderr), cph) <- streamingProcess cp
    race_ (sourceStdout $$ consumerStdout)
          (sourceStderr $$ consumerStderr)
    closeStdout
    closeStderr
    waitForStreamingProcess cph
```

which seems to pause between each process run. I assume it's blocking on either `closeStdout` or `closeStderr`.

The second way is:

```haskell
sourceProcessWithConsumers :: CreateProcess
                           -> Consumer ByteString IO () -- stdout
                           -> Consumer ByteString IO () -- stderr
                           -> IO ExitCode
sourceProcessWithConsumers cp consumerStdout consumerStderr = do
    ( ClosedStream, (sourceStdout, closeStdout)
                  , (sourceStderr, closeStderr), cph) <- streamingProcess cp
    race_ (sourceStdout $$ consumerStdout >> closeStdout)
          (sourceStderr $$ consumerStderr >> closeStderr)
    waitForStreamingProcess cph
```

This one is fast but it causes a file descriptor leak that is evident when I try to spawn 100+ concurrent processes.

Does anyone know the *right* way to do this? My knowledge of conduits is fairly limited. I would love to have a way to capture both stderr and stdout concurrently and add it to the library.

snoyberg:
Are you compiling your application with `-threaded`?

joehillen:
Turns out my slow down was something completely unrelated.

I actually get the same speed with both versions, so the first one if obviously the correct one.

Sorry for wasting your time.

But it would be nice to have a this functionality in `conduit-extra`. Without it, I had to copy this code block into my application, which is inconvenient. 

```
instance (r ~ (), MonadIO m, o ~ ByteString) => OutputSink (ConduitM i o m r) where
    osStdStream = (\(Just h) -> return $ sourceHandle h, Just CreatePipe)
[...]

PR #258, opened 2016-03-09
Add sourceProcessWithStreams

joehillen:
Relates to #250

### DIFF
--- a/conduit-extra/Data/Conduit/Process.hs
+++ b/conduit-extra/Data/Conduit/Process.hs
@@ -30,7 +30,7 @@ import Data.Conduit
 import Data.Conduit.Binary (sourceHandle, sinkHandle)
 import Data.ByteString (ByteString)
 import Control.Concurrent.Async (runConcurrently, Concurrently(..))
-import Control.Monad.Catch (MonadMask, onException, throwM)
+import Control.Monad.Catch (MonadMask, onException, throwM, finally)
 
 instance (r ~ (), MonadIO m, i ~ ByteString) => InputSource (ConduitM i o m r) where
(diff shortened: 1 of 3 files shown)


conduit-extra/Data/Conduit/Process.hs
```diff
@@ -60,6 +63,47 @@ sourceProcessWithConsumer cp consumer = do
...
+                         -> Consumer ByteString IO a -- ^stdout
+                         -> Consumer ByteString IO b -- ^stderr
+                         -> IO (ExitCode, a, b)
+sourceProcessWithStreams cp producerStdin consumerStdout consumerStderr = do
+    (  (sinkStdin, closeStdin)
+     , (sourceStdout, closeStdout)
+     , (sourceStderr, closeStderr)
+     , cph) <- streamingProcess cp
+    (_, resStdout, resStderr) <-
+      runConcurrently $ (,,)
```
snoyberg: Question: what behavior do we want if the provided `Source` or `Sink`s throw an exception? It may be appropriate to have an `onException` here that closes the streams and kills the child process. Thoughts?

joehillen: That thought occurred to me when I saw your recent addition of `withCheckedProcessCleanup`.

I think it's reasonable to do that. I suppose if a user doesn't want the subprocess killed then they can catch exceptions in the `Source`/`Sink`s.

Should that also be added to `sourceProcessWithConsumer`? I'll assume so for now and add it to the PR.

joehillen:
@snoyberg I've added the suggested exception handling.

snoyberg:
Thanks!

PR #262, opened 2016-05-04
network-conduit-tls: Expose ApplicationStartTLS

abbradar:
This should also fix documentation issues.

(See `runTCPServerStartTLS` at https://hackage.haskell.org/package/network-conduit-tls-1.2.0.1/docs/Data-Conduit-Network-TLS.html for example of current issues)


snoyberg:
Thanks!

PR #263, opened 2016-05-05
In zipConduitApp, left bias not respected mixing monadic and non-monadic conduits

farrellm:
zipConduitApp first looks for output, then evaluates monadic pipes. This results in non-monadic output always coming before monadic output, potentially violating the left bias. Monads should be evaluated first to ensure left bias is always respected.

Also adds a test that fails given the current implementation of zipConduitApp.


snoyberg:
Makes sense, good catch!

PR #264, opened 2016-05-05
Make runTLS{Client,Server}StartTLS general

abbradar:
This makes it possible to use general `runTLSClientStartTLS` without losing monadic state, and also generalizes `runTLSServerStartTLS`. Consists of several parts:

1) Make top function in `ApplicationStartTLS` return `a` -- this allows us to return and re-introduce wrapped monadic state with `liftBaseOp`.
2) Also make it parametrized by a monad. This is to not lose monadic state from the nested `startTLS` function. Another possible solution would be to use impredicative types to allow returning value from it (replace `(AppData -> IO ()) -> IO ()` with `forall a. (AppData -> IO a) -> IO a` and use ordinary `liftBaseOp`). Unfortunately, GHC doesn't work well with this.
3) Finally, generalize both `runTLS*StartTLS` functions.

Also this allows one to return value from `runTLSClientStartTLS` (not strictly needed for this patch) and drops an extra constraint from `runTLSClient`.

I'm not sure if this needs to be done some other way w.r.t. backwards compatibility.

EDIT: Also notice that (1) is also not strictly required as we already have (2) for passing state. If needed, I can split this into a different PR.

### DIFF
--- a/network-conduit-tls/Data/Conduit/Network/TLS.hs
+++ b/network-conduit-tls/Data/Conduit/Network/TLS.hs
@@ -335,12 +335,12 @@ tlsClientConfig port host = TLSClientConfig
 -- | Run an application with the given configuration.
 --
 -- Since 1.0.2
-runTLSClient :: (MonadIO m, MonadBaseControl IO m)
+runTLSClient :: (MonadBaseControl IO m)
              => TLSClientConfig
              -> (AppData -> m a)
              -> m a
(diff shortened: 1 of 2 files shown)


snoyberg:
Changing the existing type synonym breaks backwards compatibility, so I'd rather just define a new type synonym. Other than that, changes look backwards compatible.

abbradar:
Fixed!

snoyberg:
Thanks!

PR #269, opened 2016-06-25
Fix alignment restrictions

iliastsi:
The testsuite for conduit-extra fails on [armhf](https://buildd.debian.org/status/fetch.php?pkg=haskell-conduit-extra&arch=armhf&ver=1.1.13.1-1&stamp=1466081154) architectures because of unaligned memory access.

According to the docs, the peek and poke functions might require properly aligned addresses to function correctly. This is architecture dependent; thus, portable code should ensure that when peeking or poking values of some type a, the alignment constraint for a, as given by the function alignment is fulfilled.

Here I provide two patches that ensure that alignment constraints are fulfilled. The implementation requires one extra copy, but makes the code portable.

I have patched the code in two steps, first the sinkStorable implementation and then the testsuite. This hopefully demonstrates that the patches introduce no regressions.

### DIFF
--- a/conduit-extra/Data/Conduit/Binary.hs
+++ b/conduit-extra/Data/Conduit/Binary.hs
@@ -76,6 +76,10 @@ import Control.Monad.Trans.Resource (MonadResource)
 import Control.Monad.Catch (MonadThrow (..))
 import Control.Exception (Exception)
 import Data.Typeable (Typeable)
+import Foreign.Ptr (Ptr)
+#ifndef ALLOW_UNALIGNED_ACCESS
+import Foreign.Marshal (alloca, copyBytes)
+#endif
 
 -- | Stream the contents of a file as binary data.
 --
@@ -501,7 +505,14 @@ sinkStorableHelper wrap failure = do
 
     -- Given a bytestring of exactly the correct size, grab the value
     process bs = return $! wrap $! inlinePerformIO $!
-        unsafeUseAsCString bs (peek . castPtr)
+        unsafeUseAsCString bs (safePeek undefined . castPtr)
+
+    safePeek :: a -> Ptr a -> IO a
+#ifdef ALLOW_UNALIGNED_ACCESS
+    safePeek _ = peek
+#else
+    safePeek val ptr = alloca (\t -> copyBytes t ptr (sizeOf val) >> peek t)
+#endif
 {-# INLINE sinkStorableHelper #-}
 
 data SinkStorableException = SinkStorableInsufficientBytes
--- a/conduit-extra/conduit-extra.cabal
+++ b/conduit-extra/conduit-extra.cabal
@@ -33,6 +33,10 @@ Library
   if !os(windows)
       Exposed-modules: Data.Conduit.Network.Unix
 
+  if arch(x86_64) || arch(i386)
+      -- These architectures are able to perform unaligned memory accesses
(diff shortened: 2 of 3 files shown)


snoyberg:
I'm not that familiar with the code in question (I inherited it from elsewhere). Two questions:

1. I'm surprised that the test suite needed patching. Does this mean that the bytestring functions do not, in fact, allocate aligned buffers?
2. This will negatively impact performance on all platform that do not require alignment, correct? Is there any way to use conditional compilation to avoid that overhead?

snoyberg:
Correction to the previous comment, I _am_ actually the author of this code, I thought I was looking at something else. Still, original concern remains: can we avoid the performance hit of copying on systems that don't require alignment?

iliastsi:
Hi, 

> I'm surprised that the test suite needed patching. Does this mean that the bytestring functions do not, in fact, allocate aligned buffers?

In the test suite, the pokeByteOff function is used in order to fill the bytestring with a list of SomeStorable. This causes unaligned access in case, for example, the first item is of type Word8 and the second of type Double. This has nothing to do with the alignment of the bytestring itself.

> This will negatively impact performance on all platform that do not require alignment, correct? Is there any way to use conditional compilation to avoid that overhead?

I am not an expert on this, so I am not sure how bad the impact on the performance would be. From what I understand, all architectures impose alignment constraints on memory access, but some of them are able to correct an unaligned one transparently, with a performance cost. Other times, the kernel may handle the exception and fall-back to copying data (see also [unaligned-memory-access](https://www.kernel.org/doc/Documentation/unaligned-memory-access.txt) from Linux kernel).

I couldn't find a list of architectures where the above properties are described, so I thought it would be safer to enforce aligned access to all of them. If we could find such a document, then we could enable this feature only for the affected architectures, using the preprocessor.

What do you think?

snoyberg:
In our performance testing for Store, we found that on x86 architectures at least, unaligned access did not negatively hurt performance. I'd be happy to go with a whitelist approach here and just say that x86 (32- and 64-bit) continue with the code as-is, and all other architectures use the copy/align logic you've implemented here.

iliastsi:
Yes, that sounds great! Do you want me to implement the whitelisting and update the PR, or would you like to do it yourself?

I also noticed that mapM_BS uses the peek function as well, but since the testsuite now passes, I didn't change that. Do you believe we should change that as well?

P.S.: We applied the patch in Debian and the testsuite now [passes](https://buildd.debian.org/status/fetch.php?pkg=haskell-conduit-extra&arch=armhf&ver=1.1.13.1-2&stamp=1466878248) for all architectures.

snoyberg:
If you could update it, that would be great. For the `mapM_BS` function, I think alignment isn't an issue, since we're just peeking one byte, which should have an alignment of 1.

iliastsi:
Sorry for the delay, I have updated this PR.
Feel free to make any changes you like in order to match you personal coding style.

Also, I am a little worried about the use of the ``inlinePerformIO`` function. The [docs](https://github.com/haskell/bytestring/blob/master/Data/ByteString/Internal.hs#L527) seem to suggest that this function should not be used. Please consider replacing it with the ``unsafePerformIO``.

snoyberg:
Before getting too deep into `inlinePerformIO`, I'd recommend reading more about it. I actually wrote a fairly long article for the GHC wiki on the subject: https://wiki.haskell.org/Evaluation_order_and_state_tokens

PR #270, opened 2016-06-30
Export the yieldM on ConduitM.

tmcgilchrist:
Not sure the reasons why this isn't exported but I find it rather useful for code I want to keep pure and only mix into conduit at the last minute.

eg 
```haskell
deriveFeatures :: Monad m =>  Conduit Row (EitherT RowError m) OtherRow
deriveFeatures od = awaitForever (yieldM . hoistEither . deriveFeatures')

deriveFeatures' :: Row -> Either JoinOfferError OtherRow
deriveFeatures' = ....
```

### DIFF
--- a/conduit/ChangeLog.md
+++ b/conduit/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.2.7
+
+* Expose yieldM for ConduitM [#270](https://github.com/snoyberg/conduit/pull/270)
+
 ## 1.2.6.6
 
 * Fix test suite compilation on older GHCs
--- a/conduit/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/Data/Conduit/Internal/Conduit.hs
@@ -842,6 +842,9 @@ yield :: Monad m
 yield o = yieldOr o (return ())
 {-# INLINE yield #-}
 
+-- | Send a monadic value downstream for the next component to consume.
+--
+-- @since 1.2.7
 yieldM :: Monad m => m o -> ConduitM i o m ()
 yieldM mo = lift mo >>= yield
 {-# INLINE yieldM #-}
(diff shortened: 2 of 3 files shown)


snoyberg:
I'm OK exporting it, but it will need:

* A Haddock comment explaining when it's useful vs `yield`
* A `@since` line in the comment
* A mention in the ChangeLog

Could you make such changes to this PR?

tmcgilchrist:
Sure, I can do that.

After speaking to people here at work, I'm less convinced that it's a useful thing to have in my case. Most of what I need can be achieved using `mapM` and `filterM` rather than needing `yieldM` directly.

snoyberg:
FWIW, you could also write your example above as:

```haskell
awaitForever (lift . yield <=< hoistEither . deriveFeatures')
```

tmcgilchrist:
Finally got some time to add the docs, etc.
Is this right @snoyberg ?

conduit/ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## 1.2.6.7
```
snoyberg: This would be a minor version bump, not a patch, since it's not forwards-compatible (i.e., 1.2.7).

conduit/Data/Conduit/Internal/Conduit.hs
```diff
@@ -842,6 +842,9 @@ yield :: Monad m
 yield o = yieldOr o (return ())
 {-# INLINE yield #-}
 
+-- | Send a monadic value to the next component to consume.
+--
+-- Since 1.2.6.7
```
snoyberg: Could you use the newer `@since` syntax? I know most of the other docs don't do that yet...

conduit/Data/Conduit/Internal/Conduit.hs
```diff
@@ -842,6 +842,9 @@ yield :: Monad m
 yield o = yieldOr o (return ())
 {-# INLINE yield #-}
 
+-- | Send a monadic value to the next component to consume.
```
snoyberg: I think it would be worth including the word "downstream," and explaining why this is different from `yield`.

PR #274, opened 2016-07-23
Workaround for type inference regression in GHC 8.1

thomie:
`resourcet` doesn't build with GHC HEAD (ghc-8.1.20160720), due to  https://ghc.haskell.org/trac/ghc/ticket/12427.

Here is a workaround.


snoyberg:
Is there reason to believe this won't be fixed before release? I'm generally not a fan of modifying the code base to work around unreleased bugs like this, unless there's some advantage I'm not aware of.

thomie:
Fair enough. 

I ran into this while trying to reproduce another bug report with GHC HEAD, that depended on `resourcet`. But I can either use my local `resourcet` clone if that happens again, or hopefully the original GHC bug (12147) is indeed fixed by then.

PR #277, opened 2016-08-10
Remove active from code marker so it renders correct on GitHub

twhitehead:
Noticed that the *resourcet/README.md* file isn't rendering properly on GitHub due to the active keyword after the ```haskell code block marker.  The active leyword isn't used in the main *README.md* code blocks so I figure it was likely left over from pre-GitHub hosting and removing it got missed.


snoyberg:
Thanks!

PR #281, opened 2016-09-20
Add instance MonadFix m => MonadFix (ResourceT m)

mitchellwrosen:


mitchellwrosen:
I'm fairly certain this is a sensible instance, although I haven't tried proving the laws yet. It's the same code as `ReaderT`'s instance.

PR #282, opened 2016-09-30
Update the README.md tutorial

snoyberg:


bitemyapp:
Is this a typo?

```haskell
         .| mapMC magic
         .| takeWhileC (< 18)
         .| mapM_C print
```

I see `mapMC` and `mapM_C`.

bitemyapp:
The `mapM_` `mapM` distinction represented by `mapMC` and `mapM_C` might be a little confusing, but it's consistent and you highlight it in the tutorial.

snoyberg:
I was also worried about it being confusing. I'll see if I can replace one
of them with a different function, but it may not be feasible.

bitemyapp:
@snoyberg I don't think a different _function_ is needed…they're appropriate. It's not a strong objection either, it just feels like a violation of how pseudo-hungarian notation tends to work in Haskell.

You know we put most invariant arguments first, most variant arguments last for our function arguments? For suffixes/function names, I usually put the most salient/structurally significant part of the function name first, in order of decreasing change.

The other thing I would say is that I've mostly given up on mutilating function names to avoid overlap with stuff that's in base / transformers / etc. Instead I usually keep the names familiar and advise users to import qualified. The primary weakness of this IMO is that I can't re-export module aliases, a real sore spot for my coworkers and I as we use the `Import.hs` pattern from the Yesod scaffolds / Yesod stack templates.

snoyberg:
OK, I'll consider the change then for `mapM_`, but won't worry about it too much.

There _is_ a possibility of changing the syntax to using qualified import, i.e.:

```haskell
import qualified Data.Conduit.Combinator as C
import Data.Conduit.Combinator ((.|))
```

Personally I think the `import Conduit` is going to be easier for people overall, but I'm open to input on this. How's this: thumbs up this comment if you want the qualified import, and thumbs down to leave with `import Conduit` as I've done so far.

bitemyapp:
That's kind of a chicken-egg problem. Part of the problem is that new people won't use qualified imports unless they're educated to do so by tutorials/example. In work code, anything we don't keep non-overlapping in `Import.hs` and for which we aren't plucking one function, we usually use qualified imports. The only thing sparing us copious B./BL./T./TL. is your classy-prelude.

I'm not sure which way to go on this, I'd need to hammer out some concrete examples. I'm not sure we can count on beginners to use qualified imports like you say. :\

nfrisby:
I was enjoying your nice narrative when I got distracted by the following sentence.

> Notice how we've had to reimplement the behavior of `takeWhile`, `mapM`, and `mapM_` ourselves, and our logic is much harder to follow.

(I haven't read past that sentence, but I'm already convinced that this re-skinning is a huge win. Nice work! I should also mention for perspective that I was convinced of `conduit`'s benefits years ago.)

I'll lead with a suggestion: that you rewrite "our logic is much harder to follow" to "the solution is less compositional". I think that would resolve my concern. The rest of this comment will attempt to quickly motivate that.

I disagree with "our logic is much harder to follow" because that's claiming that `a` "is clearer than" `b` in the following.

```
a  = runConduit
   $ yieldMany [1..]
  .| takeC 10
  .| mapMC magic
  .| takeWhileC (< 18)
  .| mapM_C print

b = go (take 10 [1..])
  where
  go [] = return ()
  go (x:xs) = do
      y <- magic x
      when (y < 18) $ do
        print y
        go xs
```

(I used `when` here instead of the conditional. I think that's fair and makes the comparison more realistic. There are, of course, a lot of ways to skin this cat, e.g. using `foldr`.)

My main point is that "is clearer than" is unhelpfully subjective/contextual. I think `b` is pretty "clear" to begin with. That's why the sentence I quoted disrupted my reading.

What do you think of my rewording suggestion? HTH.

snoyberg:
Nice addition, agreed totally on replacing subjective phrases with objective ones.

snoyberg:
Thanks for the feedback!

PR #284, opened 2016-10-31
Remove redundant definitions of `average`

aske:
In ZipSink section of a tutorial there are two definitions of `average` function that are not used.


snoyberg:
Thanks!

PR #286, opened 2016-11-08
Use `mask_` now that 'allocate' does not use 'restore'

vdukhovni:
This should avoid potential compiler warnings.


snoyberg:
Thanks!

Issue #295, opened 2017-02-07
Possible addition to Data.Conduit.List (batch :: Int -> Conduit a m [a])


tomsmalley:
There may already be a nice way to do this that I have missed, but I found myself needing to batch items up for a downstream consumer (which was writing to a database and limited by transactions / sec). For example:

```haskell
GHCi, version 8.0.1: http://www.haskell.org/ghc/  :? for help
> runConduit $ CL.sourceList [1..17] =$ batch 5 =$ CL.iterM print =$ CL.sinkNull
[1,2,3,4,5]
[6,7,8,9,10]
[11,12,13,14,15]
[16,17]
```

Would others find this useful or is it too specialised to add?

I implemented it by a minor modification to `groupBy`

```haskell
batch :: Monad m => Int -> Conduit a m [a]
batch n =
    start
  where
    start = await >>= maybe (return ()) (loop n id)

    loop count rest x =
        await >>= maybe (yield (x : rest [])) go
      where
        go y
            | count > 1 = loop (count - 1) (rest . (y:)) x
            | otherwise = yield (x : rest []) >> loop n id y
```

snoyberg:
Seems like a good addition to me. Looks like a name more in line with the
rest of the ecosystem may be chunksOf:

https://www.stackage.org/lts-7.19/hoogle?q=chunksOf

PR certainly welcome.

tomsmalley:
Well spotted, I submitted a PR (#296) but it doesn't include a chunksOfS version.

PR #296, opened 2017-02-07
Add function Data.Conduit.List.chunksOf (issue #295)

tomsmalley:
Introduced an additional dependency in the test suite.
Stream version not included.
See issue #295

### DIFF
--- a/conduit/Data/Conduit/List.hs
+++ b/conduit/Data/Conduit/List.hs
@@ -638,19 +638,19 @@ STREAMING0(consume, consumeC, consumeS)
 --
 -- Subject to fusion
 --
--- Since ???
+-- Since 1.2.9
 chunksOf :: Monad m => Int -> Conduit a m [a]
 chunksOf n =
     start
   where
-    start = await >>= maybe (return ()) (loop n id)
+    start = await >>= maybe (return ()) (\x -> loop n (x:))
 
-    loop count rest x =
-        await >>= maybe (yield (x : rest [])) go
+    loop !count rest =
+        await >>= maybe (yield (rest [])) go
       where
         go y
-            | count > 1 = loop (count - 1) (rest . (y:)) x
-            | otherwise = yield (x : rest []) >> loop n id y
+            | count > 1 = loop (count - 1) (rest . (y:))
+            | otherwise = yield (rest []) >> loop n (y:)
 
 -- | Grouping input according to an equality function.
 --
--- a/conduit/Data/Conduit/List.hs
+++ b/conduit/Data/Conduit/List.hs
@@ -50,6 +50,7 @@ module Data.Conduit.List
     , scanl
     , scan
     , mapAccum
+    , chunksOf
     , groupBy
     , groupOn1
     , isolate
@@ -80,6 +81,7 @@ import Prelude
     , Enum, Eq
     , maybe
     , (<=)
+    , (>)
     )
 import Data.Monoid (Monoid, mempty, mappend)
 import qualified Data.Foldable as F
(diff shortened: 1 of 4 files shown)


conduit/Data/Conduit/List.hs
```diff
@@ -631,6 +633,25 @@ consumeC =
 {-# INLINE consumeC #-}
 STREAMING0(consume, consumeC, consumeS)
 
+-- | Group a stream into chunks of a given size. The last chunk may contain
+-- fewer than n elements.
+--
+-- Subject to fusion
+--
+-- Since ???
```
snoyberg: Can you put this at 1.2.9, and adjust the .cabal file and ChangeLog accordingly?

conduit/Data/Conduit/List.hs
```diff
@@ -631,6 +633,25 @@ consumeC =
...
+chunksOf :: Monad m => Int -> Conduit a m [a]
+chunksOf n =
+    start
+  where
+    start = await >>= maybe (return ()) (loop n id)
+
+    loop count rest x =
```
snoyberg: May be worthwhile to use a bang pattern on `count` (i.e., `!count`) to avoid laziness. The guard below should be enough for the strictness analyzer, but better safe than sorry.

conduit/Data/Conduit/List.hs
```diff
@@ -631,6 +633,25 @@ consumeC =
...
+-- fewer than n elements.
+--
+-- Subject to fusion
+--
+-- Since ???
+chunksOf :: Monad m => Int -> Conduit a m [a]
+chunksOf n =
+    start
+  where
+    start = await >>= maybe (return ()) (loop n id)
```
snoyberg: Instead of passing the `x` around below, how about:

```haskell
start = await >>= maybe (return ()) (\x -> loop n (x:))
```

You _could_ point-free that... but please don't, it'll hurt my head :)

conduit/Data/Conduit/List.hs
```diff
@@ -631,6 +633,25 @@ consumeC =
...
+
+    loop count rest x =
+        await >>= maybe (yield (x : rest [])) go
+      where
+        go y
+            | count > 1 = loop (count - 1) (rest . (y:)) x
```
snoyberg: Very nice way to avoid the negative-input case 👍

conduit/test/main.hs
```diff
@@ -235,6 +236,20 @@ main = hspec $ do
                     C.=$ CL.fold (+) 0
             x `shouldBe` 2 * sum [1..10 :: Int]
 
+        it "chunksOf" $ do
+            let input = [1..16] :: [Int]
+            x <- runResourceT $ CL.sourceList input
```
snoyberg: I think the `runResourceT` is unnecessary here.

conduit/test/main.hs
```diff
@@ -235,6 +236,20 @@ main = hspec $ do
                     C.=$ CL.fold (+) 0
             x `shouldBe` 2 * sum [1..10 :: Int]
 
+        it "chunksOf" $ do
+            let input = [1..16] :: [Int]
```
snoyberg: How about turning this into a `prop` with a variable size? Or add that as an additional case?

conduit/test/main.hs
```diff
@@ -235,6 +236,20 @@ main = hspec $ do
...
+            x <- runResourceT $ CL.sourceList input
+                    C.$$ CL.chunksOf 5
+                    C.=$ CL.consume
+            x `shouldBe` DLS.chunksOf 5 input
+
+        it "chunksOf (negative)" $ do
+            let input = [1..16] :: [Int]
+            x <- runResourceT $ CL.sourceList input
```
snoyberg: Also not needed here

snoyberg:
Looks good!

tomsmalley:
Changes implemented!
Let me know if I'm missing something.

snoyberg:
Looks great, thanks!

PR #307, opened 2017-04-19
Deprecated older synonyms and operators (fixes #283)

### LINKED ISSUE #283: RFC: Stop using the type synonyms in library type signatures
Reasons:

* Adds extra things for people to learn
* Difference between `Source` and `Producer` (or `Sink` and `Consumer`) is subtle and tricky
* Relying on a poorly supported language extension

Downsides:

* Makes the type signatures more confusing to read (arguable)
* Possibly breaks lots of old docs

Please use 👍 and 👎 reactions on this description to express support/opposition to this change.

snoyberg:
This implements the least painful bits of #283: deprecating old stuff without changing the types or removing anything. I believe that the new operators and functions have been present long enough now that introducing a deprecation is acceptable. Issues of getting rid of the old synonyms, replacing `ConduitM` with `Conduit`, or playing with `Void` vs `()` vs free variables are (intentionally) not addressed here.


snoyberg:
Handled by #338

PR #308, opened 2017-04-28
Update to work with conduit-combinators 1.1.0

AlainODea:
- Handle type signature changes (stdoutC consumes ByteString as of 1.1.0)

 - Use stack script to detect missing packages

 - Add missing packages to stack arguments

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ with.
 
 ``` haskell
 #!/usr/bin/env stack
--- stack --resolver lts-6.19 runghc --package conduit-combinators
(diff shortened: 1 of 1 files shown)


AlainODea:
My pleasure, Michael. Conduit is a great framework and fun to work with. This tutorial was instrumental to me getting into it and that one small bump with **stdoutC** made me realize I had an easy fix and an opportunity to give back.

snoyberg:
Doc fixes like this are _always_ welcome, especially thorough ones like
this! Again, much appreciated.

PR #316, opened 2017-06-09
Add unfoldEither and unfoldEitherM functions

ivan-m:
This provides the ability to produce a Producer-like Conduit that has
a return value.

Not sure if it's worth defining a new type alias to make the `forall i` explicit rather than implicit.

I have a use case for this where I am converting a streaming source from another library to a Conduit and I would like to keep the return value rather than just discarding it.

### DIFF
--- a/conduit/ChangeLog.md
+++ b/conduit/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.2.11
+
+* Add `unfoldEither` and `unfoldEitherM` to `Data.Conduit.List`
+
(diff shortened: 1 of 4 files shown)


snoyberg:
Thanks!

ivan-m:
Ta!

I'm guessing the lack of type aliases is OK considering #283 ?

snoyberg:
Yes, we'll hopefully deprecate them soon and in the far future remove them.

PR #327, opened 2017-09-08
Added a strictness footnote.

reactormonk:
There's probably a better way to do this, do you have an idea?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ consuming that data.
 
 Some of the reasons you'd like to use conduit are:
 
-* Constant memory usage over large data
+* Constant memory usage over large data<sup>[1](#strictness)</sup>
 * Deterministic resource usage (e.g., promptly close file handles)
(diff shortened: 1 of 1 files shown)


snoyberg:
I think this comment is too vague. I'd rather point to another source on the difference between WHNF and NF. The conversation we've been having has encouraged me to write a blog post on the subject. Maybe that will help clear things up (and be a good source to link to).

reactormonk:
Would it be overkill/wrong approach to add `NFData` to most of the conduit functions?

snoyberg:
It's conceivable to add a set of NFData functions, but it wouldn't be a
replacement for the existing functions. NFData introduces both an extra
constraint, and extra overhead. Using strict data types is usually the
right answer in these cases.

snoyberg:
Closing this for now due to the concerns I cited. If you have a more moderate wording, a new PR is welcome.

PR #328, opened 2017-09-21
Fix pass in ConduitM MonadWriter instance

ivan-timokhin:
Previous implementation was equivalent to `fmap fst`, which is
probably not the intended behavior.

As an example, this

```haskell
execWriter $ runConduit $ pass $ tell [1] >> pure ((), (2:))
```

previously evaluated to `[1]`, and now evaluates to `[2,1]`.

### DIFF
--- a/conduit/ChangeLog.md
+++ b/conduit/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.2.12.1
+
+* Fix `pass` in `ConduitM` `MonadWriter` instance
+
 ## 1.2.12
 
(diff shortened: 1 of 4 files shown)


snoyberg:
Thanks for the fix, I'm honestly not that familiar with the `pass` method. Do you think it would be possible to add a test case for proper behavior so that this doesn't get accidentally broken again in the future?

ivan-timokhin:
Done. Anything I can/should do about that timeout in Travis?

snoyberg:
No, nothing to be done for the timeouts, except restarting the test. Thanks for the added test case and the changelog update as well, much appreciated.

PR #329, opened 2017-09-26
Remove parameter to ResIO type alias

davidspies:
This allows the use of ResIO as an argument to some other Monad Transformer.

eg.
ReaderT Int ResIO x


snoyberg:
Looks good, would you mind making a minor version bump and a note in the ChangeLog?

Issue #334, opened 2017-11-23
streamingProcess and stdin buffering


CthulhuDen:
With default Handle buffering mode it's not possible to interact with programs because input gets buffered.

This may seem like an issue with `streaming-commons`, but it's only an issue when using Conduit as input sink for a new process, because then we have no access to underlying handle and so cannot change buffering mode.

CthulhuDen:
This was probably a collateral damage from fixing https://github.com/snoyberg/conduit/issues/322.

snoyberg:
Shoot, this looks like a "damned if you do, damned if you don't" situation. I have a few different thoughts on how to address this.

1. Instead of getting a `Sink`, just ask for the raw `Handle`, modify it as desired, and then use `sinkHandle`. That should work immediately, though it's not as elegant from a user perspective.
2. This requires a breaking change and losing some niceness around type inference. The idea is to allow for a `Flush ByteString` and `Builder` instance for `InputSource`. Then you'd be able to `yield Flush` to force flushing.
3. Add some special logic to treat `yield B.empty` as a command to flush.
4. Add some newtypes to state explicitly that you want the stream to have different buffering strategies.

Do you have any thoughts on these options?

CthulhuDen:
Only issue with your proposed addition is I think it need third constructor which will pass chunk and flush, so that user can just plug 'mapC FlushChunk` or something before the sink.

Actually, I think it's generally a good idea to flush immediately into process stdin, so I was thinking about making this default behaviour, and maybe add Newtype to override. Currently, I use the following newtype as a workaround, but it could be done in reverse:

```haskell
newtype FlushingSink o m = Flushing (ConduitM BS.ByteString o m ())

sinkFlushingHandle :: MonadIO m => Handle -> ConduitM BS.ByteString o m ()
sinkFlushingHandle h = awaitForever $ \bs -> liftIO $ BS.hPut h bs >> hFlush h

instance MonadIO m => InputSource (FlushingSink o m) where
  isStdStream = (\(Just h) -> return $ Flushing $ sinkFlushingHandle h, Just CreatePipe)

instance (MonadIO m, MonadIO n) => InputSource (FlushingSink o m, n ()) where
  isStdStream = (\(Just h) -> return (Flushing $ sinkFlushingHandle h, liftIO $ hClose h), Just CreatePipe)
```

CthulhuDen:
[...]

PR #336, opened 2017-12-11
Add Handle sinks with buffering control, define more Process stdin instances

CthulhuDen:
Re: #334, upon more consideration I copied your (@snoyberg) code, just added comments and instances. Not sure if this need mentioning in the docs (PROCESS.md), as I now realize that this [expect][1]-like usage is really uncommon case for conduit.

[1]: https://linux.die.net/man/1/expect

### DIFF
--- a/conduit-extra/Data/Conduit/Binary.hs
+++ b/conduit-extra/Data/Conduit/Binary.hs
@@ -28,6 +28,8 @@ module Data.Conduit.Binary
     , sinkSystemTempFile
     , sinkHandle
     , sinkIOHandle
+    , sinkHandleBuilder
+    , sinkHandleFlush
     , withSinkFile
     , withSinkFileBuilder
       -- ** Conduits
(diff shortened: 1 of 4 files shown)


conduit-extra/Data/Conduit/Binary.hs
```diff
@@ -167,14 +170,31 @@ sourceIOHandle :: MonadResource m
...
 -- Since 0.3.0
 sinkHandle :: MonadIO m
            => IO.Handle
            -> Consumer S.ByteString m ()
 sinkHandle h = awaitForever (liftIO . S.hPut h)
 
+-- | Stream incoming builders, executing them directly on the given 'IO.Handle'
+-- buffer. Note that this function does /not/ automatically close the @Handle@
+-- when processing completes. Pass 'Data.ByteString.Builder.Extra.flush'
+-- to flush buffer.
```
snoyberg: Can you provide a minor version bump in the cabal file, update the `ChangeLog`, and add `@since` comments to all new identifiers and instances?

CthulhuDen: Done.

snoyberg: Can you also include the changelog update? And it's not necessary, but for future PRs it's better to use the new `@since` syntax for Haddock comments.

CthulhuDen: Right, forgot about changelog

CthulhuDen: I've addressed the points and rebased.

snoyberg:
Thanks!

PR #338, opened 2017-12-20
Big breaking change :) See description

snoyberg:
* Finally moves ahead with the deprecations described in #283 
* Renames `ConduitM` to `ConduitT`
* Drops usage of `monad-control` wherever possible, instead favoring `unliftio`
* Still maintains dependencies on `exceptions` due to usage of `MonadThrow`, but that's up for discussion

This PR is _not_ ready for merge, I suspect there will be feedback as well as some more cleanups to throw into the mix.


Ericson2314:
I was hoping there'd still be a monomorphic way to `hoist` a `ResourceT` after this, but that does not appear to be the case?

snoyberg:
I believe transPipe (https://www.stackage.org/haddock/lts-12.0/conduit-1.3.0.3/Data-Conduit.html#v:transPipe) should fulfill that role.

Ericson2314:
Thank you. Indeed `transPipe` and [`transResourceT`](https://www.stackage.org/haddock/lts-12.0/resourcet-1.2.1/Control-Monad-Trans-Resource.html#v:transResourceT) do just that.

PR #342, opened 2017-12-27
Non-quadratic lines (fixes #341)

### LINKED ISSUE #341: lines suspected O(n²)
I suspect that https://github.com/snoyberg/conduit/blob/33feb2d32da4a0564d01b6c61481b7e5f8084597/conduit-extra/Data/Conduit/Text.hs#L87-L101 is O(n²) if I send it one char after the other.

Ideas:

* Switch to `Text.Lazy` for `T.append` to avoid quadratic append
* Don't search (T.`break`) from the beginning each time, only searc for `\n` in the newly `await`ed chunk

snoyberg:
Nice catch @nh2. I'd rather use a difference list than lazy Text, as the
latter will still have some quadratic complexity. Does this address the
case you were concerned with?

### DIFF
--- a/conduit-extra/ChangeLog.md
+++ b/conduit-extra/ChangeLog.md
@@ -1,3 +1,8 @@
+## 1.2.3.2
+
+* Avoid quadratic complexity in `lines`
+  [#341](https://github.com/snoyberg/conduit/issues/341)
+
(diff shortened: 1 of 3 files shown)


snoyberg:
Added to #338

PR #345, opened 2018-01-05
Add Semigroup instances for Pipe and ConduitM

RyanGlScott:
Currently, `conduit-1.2.12.1` fails to build on GHC 8.4.1 since `Semigroup` has become a superclass of `Monoid`, and `Pipe` and `ConduitM` lack `Semigroup` instances. This fixes the issue.

I opted to avoid some CPP by depending on the `semigroups` package to provide the `Semigroup` class on GHC 7.10 and older (before `Semigroup` was added to `base`). If you'd prefer to avoid incurring an extra dependency, I could alternatively remove the `semigroups` dependency and only define the `Semigroup` instances on `base-4.9` or later (at the cost of some CPP).


RyanGlScott:
OK, the Travis build now passes (save for one matrix entry which stalled out).

snoyberg:
Thanks! I'll merge as-is, but add some cabal file conditionals to only use the semigroups dependency in older GHCs.

Issue #346, opened 2018-01-05
runResourceT doens't rethrow exceptions in finalizers


nh2:
```
runResourceT $ allocate (return 'x') (\_x -> putStrLn "finalizer running" >> error "erroring")
```

prints only

```
finalizer running
```

and swallows the `error`.

It would be nice to have a variant of `runResourceT` that rethrows errors happening in finalizers.

snoyberg:
Fixed in resourcet 1.2.

PR #347, opened 2018-01-07
runResourceTChecked #346

snoyberg:
Pinging @nh2

### DIFF
--- a/resourcet/Control/Monad/Trans/Resource/Internal.hs
+++ b/resourcet/Control/Monad/Trans/Resource/Internal.hs
@@ -373,12 +373,34 @@ registerType istate rel = I.atomicModifyIORef istate $ \rm ->
 -- exception during cleanup.
 --
 -- @since 1.1.11
-data ResourceCleanupException = ResourceCleanupException !SomeException ![SomeException]
+data ResourceCleanupException = ResourceCleanupException
+  { rceOriginalException :: !(Maybe SomeException)
+  -- ^ If the 'ResourceT' block exited due to an exception, this is
+  -- that exception.
+  --
+  -- @since 1.1.11
+  , rceFirstCleanupException :: !SomeException
+  -- ^ The first cleanup exception. We keep this separate from
+  -- 'rceOtherCleanupExceptions' to prove that there's at least one
+  -- (i.e., a non-empty list).
+  --
+  -- @since 1.1.11
+  , rceOtherCleanupExceptions :: ![SomeException]
+  -- ^ All other exceptions in cleanups.
+  --
+  -- @since 1.1.11
+  }
   deriving (Show, Typeable)
 instance Exception ResourceCleanupException
 
-stateCleanupChecked :: ReleaseType -> I.IORef ReleaseMap -> IO ()
-stateCleanupChecked rtype istate = E.mask_ $ do
+-- | Clean up a release map, but throw a 'ResourceCleanupException' if
+-- anything goes wrong in the cleanup handlers.
+--
+-- @since 1.1.11
+stateCleanupChecked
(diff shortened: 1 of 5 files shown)


resourcet/Control/Monad/Trans/Resource/Internal.hs
```diff
@@ -366,3 +368,44 @@ registerType istate rel = I.atomicModifyIORef istate $ \rm ->
...
+stateCleanupChecked :: ReleaseType -> I.IORef ReleaseMap -> IO ()
+stateCleanupChecked rtype istate = E.mask_ $ do
+    mm <- I.atomicModifyIORef istate $ \rm ->
+        case rm of
+            ReleaseMap nk rf m ->
+                let rf' = rf - 1
+                 in if rf' == minBound
+                        then (ReleaseMapClosed, Just m)
+                        else (ReleaseMap nk rf' m, Nothing)
+            ReleaseMapClosed -> throw $ InvalidAccess "stateCleanup"
```
nh2: Should probably be `InvalidAccess "stateCleanupChecked"`

resourcet/Control/Monad/Trans/Resource/Internal.hs
```diff
@@ -366,3 +368,44 @@ registerType istate rel = I.atomicModifyIORef istate $ \rm ->
...
+-- Note that this returns values in reverse order, which is what we
+-- want in the specific case of this function.
+mapMaybeM :: Monad m => (a -> m (Maybe b)) -> [a] -> m [b]
+mapMaybeM f =
+    go []
+  where
+    go bs [] = return bs
+    go bs (a:as) = do
+      mb <- f a
+      let bs' = maybe id (:) mb bs
```
nh2: Bang `!bs'` or `go !bs` to avoid `id (id (...))`?

resourcet/Control/Monad/Trans/Resource/Internal.hs
```diff
@@ -366,3 +368,44 @@ registerType istate rel = I.atomicModifyIORef istate $ \rm ->
...
+            case res of
+                [] -> return () -- nothing went wrong
+                e:es -> E.throwIO $ ResourceCleanupException e es
+        Nothing -> return ()
+  where
+    try :: IO () -> IO (Maybe SomeException)
+    try io = fmap (either Just (\() -> Nothing)) (E.try io)
+
+-- Note that this returns values in reverse order, which is what we
+-- want in the specific case of this function.
```
nh2: Nitpick: Could name it `mapMaybeReverseM` then, I was thinking about exactly the order when reading the name at the call site.

snoyberg:
Good catches, I especially liked the `!bs` catch, I'd not thought of that. I originally did explicit pattern matching, and reverted to that version to avoid the `id` calls.

nh2:
> will test it right now on my project

That is done now, it works well. 👍 

I guess it could be debated on whether there should be ways to determine whether it should throw as soon as it hits the first exception in a registered finalizer, or after it tries to run all of them as it does not.
An argument for having a way to do the former is that there might be cases where having a finalizer fail makes _any_ subsequent code unsafe to run, also other finalizers. This was the case e.g. in https://github.com/serokell/rocksdb-haskell/pull/10, where calling `close` on a database twice would have te second one segfault. However, I don't need the only-one-finalizer-exception approach right now, as I could turn that segfault into an exception, and so it doesn't matter if I get one or multiple of those (for a segfault it would).

snoyberg:
I'd be very hesitant about adding a only-first function, since it violates the primary purpose of `ResourceT`: ensuring that all cleanup functions are run.

PR #348, opened 2018-01-30
Add warnings about some consumers (drop*, mapM*)

neongreen:
See https://github.com/snoyberg/mono-traversable/pull/160.


snoyberg:
Awesome, thanks!

PR #355, opened 2018-03-05
Improve takeWhile(E) by making them tail recursive

luispedro:
Previously, they were not syntactically tail recursive as they jumped
between the `loop` and the `go` functions. Thus, thunks would
accumulate.

I debugged a large memory usage issue in [ngless](http://ngless.embl.de) down to this little fellow, `takeWhile`. This new version cuts memory usage by 90% (from ~250GB down to 25). 

(I tested it on ngless, but let it run the automated testing on travis before merging).

### DIFF
--- a/conduit/src/Data/Conduit/Combinators.hs
+++ b/conduit/src/Data/Conduit/Combinators.hs
@@ -1599,10 +1599,7 @@ takeWhile :: Monad m
 takeWhile f =
     loop
   where
-    loop = await >>= maybe (return ()) go
-    go x = if f x
-        then yield x >> loop
-        else leftover x
+    loop = await >>= maybe (return ()) (\x -> if f x then yield x >> loop else leftover x)
 {-# INLINE takeWhile #-}
 
 -- | Stream all elements downstream that match the given predicate in a chunked stream.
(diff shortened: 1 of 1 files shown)


snoyberg:
Unless I'm misreading this, the change is simply manually inlining a helper function, correct? Would you be able to test if instead adding `INLINE` pragmas has the same effect?

luispedro:
Yes, it's just a syntactical manipulation. I vaguely remember reading once that these tail-recursion loops (`loop` calls `go` tail-recursively which calls `loop` tail-recursively) can mess up GHC's optimizer, but I cannot find the documentation now (and could be misremembering). Empirically, this small change (with `-O2`) made the memory profile go from linearly increasing with the input to constant.

I can test a `INLINE`s, but what exactly did you have in mind?

snoyberg:
I was thinking of trying an `INLINE` on the `go` functions. Due to the mutually recursive nature, I'm not sure that it will make any difference.

luispedro:
I'm going to close this as I couldn't reproduce the benefits on a smaller test set. It seems that there was something in my code that triggered GHC's optimizations just right when it was done like this, but out of that whole context, this PR does nothing.

luispedro:
After a lot more debugging, I realized I was getting fooled by the compiler:

The problem **only shows up when using `-O2`**! So, whenever, I was running the profiling builds, I didn't see it.

I put up a stripped down example: https://github.com/luispedro/TestingNGLESS

PR #357, opened 2018-03-08
Documentation: Point out equality of fuse, (.|) and (=$=).

### LINKED ISSUE #356: Docs: confusion about (.|), fuse, (=$=) and depreciation
The documentation for `(.|)` seems to suggest that it is the same as `fuse`. However, the documentation for `fuse` says that it is a synonym for `(=$=)`. As a first time user I conclude that `(.|) = (=$=)`, and looking at the source code they indeed seem to be equivalent. However, this is not obvious just from the documentation.

**Suggestion 1**: If these functions are the same, their documentation should say so unambiguously. Currently, the function comments for `(.|)` and `(=$=)` have slightly different wordings which could suggest that they are not.

The source for `(=$=)` says `{-# DEPRECATED (=$=) "Use .|" #-}`, suggesting that `(.|)` is preferred.

**Suggestion 2**: The stackage haddocks should include the remark that (=$=) is deprecated.

This applies to [conduit-1.2.13.1](https://www.stackage.org/haddock/lts-10.8/conduit-1.2.13.1/Data-Conduit.html#v:-61--36--61-) (in LTS-10.8).

limick:
Also, fix dangling where clause that prevented depreciation remark concerning (=$=) from being included in haddock documentation. Fixes #356.

### DIFF
--- a/conduit/src/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/src/Data/Conduit/Internal/Conduit.hs
@@ -724,7 +724,7 @@ fuse = (=$=)
 --
 -- Output from the upstream (left) conduit will be fed into the
 -- downstream (right) conduit. Processing will terminate when
--- downstream (right) returns, and both @Conduit@s will be closed.
+-- downstream (right) returns.
 -- Leftover data returned from the right @Conduit@ will be discarded.
 --
 -- Equivalent to 'fuse' and '=$=', however the latter is deprecated and will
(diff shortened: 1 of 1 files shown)


conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -721,8 +724,11 @@ fuse = (=$=)
 --
 -- Output from the upstream (left) conduit will be fed into the
 -- downstream (right) conduit. Processing will terminate when
--- downstream (right) returns. Leftover data returned from the right
--- @Conduit@ will be discarded.
+-- downstream (right) returns, and both @Conduit@s will be closed.
```
snoyberg: The "closed" comment is no longer relevant in 1.3 due to the removal of finalizers.

conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -763,12 +769,15 @@ src $$ sink = do
 {-# RULES "conduit: =$ is =$=" (=$) = (=$=) #-}
 {-# DEPRECATED (=$) "Use .|" #-}
 
--- | Fusion operator, combining two @Conduit@s together into a new @Conduit@.
---
--- Both @Conduit@s will be closed when the newly-created @Conduit@ is closed.
+-- | Combine two @Conduit@s together into a new @Conduit@ (aka 'fuse').
```
snoyberg: Given that this is already deprecated, how about just documenting as "deprecated operator for `.|`"

limick: Since this is what the depreciation remark already says, I've changed the wording to "Deprecated fusion operator". Together with the depreciation remark that should give a clear picture.

snoyberg:
Thanks!

PR #366, opened 2018-04-09
resourcet: Support exceptions-0.10

### LINKED ISSUE #365: MonadMask ResourceT lacks an implementation of generalBracket
This leads to a runtime crash.

ocharles:
Fixes #365

I made the decision to act like a reader monad transformer here, but that might not be correct. Maybe you want to roll back changes to the `IORef` - I'll have to defer to someone who understands this stuff more.

### DIFF
--- a/resourcet/ChangeLog.md
+++ b/resourcet/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.2.1
+
+* Support `exceptions-0.10`.
+
 ## 1.2.0
 
 * Drop `monad-control` and `mmorph` dependencies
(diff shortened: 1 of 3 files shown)


ocharles:
Ok, let me see what I can do.

snoyberg:
Actually, I should be able to patch this up.

snoyberg:
Thanks for the PR, I've added back 0.8 support (skipping 0.9, since it's deprecated), added an upper bound, and added a test case for this kind of bug.

ocharles:
Thanks!

Issue #371, opened 2018-05-05
Mark ResourceT as Trustworthy?


nicuveo:
ResourceT is currently unsupported by Control.Monad.Base, but due to resourceT not being compiled as safe (or at least trustworthy), it is currently not possible to fix Control.Monad.Base.

nicuveo:
Sorry, I should have been more clear! Truth be told, I'm new to Safe Haskell.

Basically, I want to add support for ResourceT in Control.Monad.Base, which should be a one-liner, except for the fact that Control.Monad.Base is built with -XSafe: it only agrees to import modules that are considered Safe, or at least Trustworthy. And sadly, ResourceT is neither.

If my understanding of both this Safe model and of ResourceT is correct, it shouldn't be hard for ResourceT to be dimmed trustworthy. This shouldn't have an impact, since trustworthy modules are not limited in what they can import (again, assuming my understanding is correct).

I forked the repo, I'll have a quick look to see if it's doable!

nicuveo:
I had a look: most of resourcet can actually be marked as Safe. The only module that can only be Trustworthy is Control.Monad.Resource.Internal, due to Control.Monad.Primitive itself not being safe.

If you don't want to take the risk of making all those modules safe (which, indeed, might break if dependencies stop being safe or trustworthy), I can do instead a patch that marks them trustworthy. The only downside is that it'll add a few warnings about how the module is marked as trustworthy but is actually safe.

nicuveo:
Yeah... in parallel I'm going to try to convince the maintainer of transformers-base to remove -XSafe, or at least perhaps to put it behind a flag. In the meantime I'll prepare the pull request that makes resourcet trustworthy, not safe, which AFAIU should be harmless (famous last words).

nicuveo:
I am an idiot, and didn't think of the [obvious solution](https://github.com/snoyberg/conduit/pull/381).

snoyberg:
Not going to be doing this, closing.

PR #372, opened 2018-05-06
Marked resourcet as Trustworthy.

nicuveo:
In order to be imported from Safe Haskell modules, resourcet needs to be at
least deemed "Trustworthy". In its current state, most modules could be marked
as "Safe", except for resourcet/Control/Monad/Trans/Resource/Internal.hs, which
relies on Control.Monad.Primitive, which is itself not safe.

Instead of trying to make everything "Safe", this patch only marks resourcet as
"Trustworthy", which puts no restriction upon what it can do or import, and
should therefore be a completely harmless change. It only causes GHC to emit a
warning, since most of the files marked as "Trustworhty" are actually deemed
"Safe" by the compiler, which is why this patch also adds
"-fno-warn-trustworthy-safe" to the compilation flags.

Safe Haskell documentation:
https://downloads.haskell.org/~ghc/8.2.2/docs/html/users_guide/safe_haskell.html

Discussion on this change:
https://github.com/snoyberg/conduit/issues/371

### DIFF
--- a/resourcet/Control/Monad/Trans/Resource.hs
+++ b/resourcet/Control/Monad/Trans/Resource.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE Trustworthy #-}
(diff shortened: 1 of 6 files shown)


snoyberg:
My previous statements still remain: I'm not very comfortable adding any pragmas for Safe Haskell.

PR #380, opened 2018-06-25
network-conduit-tls.cabal: Drop SSL3 from supported versions

erikd:
SSL3 is vulnerable to the Poodle (Padding Oracle On Downgraded Legacy
Encryption) attack. See CVE-2014-3566.

Also bump cabal version and add a ChangeLog entry.

### DIFF
--- a/network-conduit-tls/ChangeLog.md
+++ b/network-conduit-tls/ChangeLog.md
@@ -1,3 +1,7 @@
+## 1.3.2
+
+* Use the recommended SSL versions from TLS
+
 ## 1.3.1
 
 * Use the default ciphersuite from TLS
--- a/network-conduit-tls/Data/Conduit/Network/TLS.hs
+++ b/network-conduit-tls/Data/Conduit/Network/TLS.hs
@@ -131,7 +131,6 @@ serverHandshake socket creds = do
         { TLS.serverWantClientCert = False
         , TLS.serverSupported = def
             { TLS.supportedCiphers = TLSExtra.ciphersuite_default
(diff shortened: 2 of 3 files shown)


network-conduit-tls/Data/Conduit/Network/TLS.hs
```diff
@@ -131,7 +131,7 @@ serverHandshake socket creds = do
         { TLS.serverWantClientCert = False
         , TLS.serverSupported = def
             { TLS.supportedCiphers = TLSExtra.ciphersuite_default
-            , TLS.supportedVersions = [TLS.SSL3,TLS.TLS10,TLS.TLS11,TLS.TLS12]
+            , TLS.supportedVersions = [TLS.TLS10,TLS.TLS11,TLS.TLS12]
```
vdukhovni: My suggestion was not to remove `TLS.SSL3` and leave the rest. It was to completely leave out the `TLS.supportedVersions` override and let the underlying TLS library (`hs-tls`) provide  a sensible default.  Which it does....

erikd: I did like this because it is explicit, but yes, its probably better if TLS decides on the default.

network-conduit-tls/ChangeLog.md
```diff
@@ -1,3 +1,7 @@
+## 1.3.2
+
+* Drop SSLv3 from supported versions.
```
snoyberg: How about clarifying that this simply uses the default upstream value, since this comment may fall out of sync in the future with changes to the `tls` library.

erikd: Done!

PR #381, opened 2018-07-15
Added MonadBase instance for ResourceT.

nicuveo:
My [previous attempt](https://github.com/snoyberg/conduit/pull/372) was to try to make ResourceT compatible with Safe Haskell, so that I could add that instance in Control.Monad.Base. However, as Control.Monad.Base's maintainer [pointed out](https://github.com/mvv/transformers-base/issues/16), it makes more sense to simply add that instance in ResourceT.

I don't know why I didn't think of that in the first place. ^^'


snoyberg:
I don't intend to add the `transformers-base` dependency back.

nicuveo:
Do you mind if I ask why?

nicuveo:
I understand not wanting the dependency; but since neither side wants it, we're left with ResourceT being incompatible with MonadBase. :/

PR #383, opened 2018-08-02
Add section to document combinator letters

roman:
After reading [your blogpost of Haskell vs Rust for streaming](https://www.fpcomplete.com/blog/2018/07/streaming-utf8-haskell-rust), I noticed the way you fleshed out the rationale for the letters on the combinators was really clear, and that I didn't find something like that in the haddock documentation before.

This pull request adds a new section to the `Data.Conduit.Combinators` module called "Naming Scheme" and adds a link to it on the documentation of all the functions that have special letters in their name.

### DIFF
--- a/conduit/src/Data/Conduit/Combinators.hs
+++ b/conduit/src/Data/Conduit/Combinators.hs
@@ -7,21 +7,18 @@
 {-# LANGUAGE NoImplicitPrelude         #-}
 {-# LANGUAGE NoMonomorphismRestriction #-}
 {-# LANGUAGE BangPatterns #-}
--- | This module is meant as a replacement for Data.Conduit.List.
+
+-- | This module is meant as a replacement for 'Data.Conduit.List'.
 -- That module follows a naming scheme which was originally inspired
 -- by its enumerator roots. This module is meant to introduce a naming
 -- scheme which encourages conduit best practices.
 --
--- There are two versions of functions in this module. Those with a trailing
--- E work in the individual elements of a chunk of data, e.g., the bytes of
--- a ByteString, the Chars of a Text, or the Ints of a Vector Int. Those
--- without a trailing E work on unchunked streams.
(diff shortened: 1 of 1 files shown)


conduit/src/Data/Conduit/Combinators.hs
```diff
@@ -304,6 +301,21 @@ import           Data.Sequences (LazySequence (..))
 --   so there's no way for the conduit to guarantee a certain amount
 --   of demand from the upstream.
 
+-- $naming_scheme
+--
+-- [@Functions ending with C@]: This postfix indicates a function returns a
```
snoyberg: Note that this applies to the `Conduit` module, but _not_ `Data.Conduit.Combinators`. That should probably be clarified here.

conduit/src/Data/Conduit/Combinators.hs
```diff
@@ -304,6 +301,21 @@ import           Data.Sequences (LazySequence (..))
...
+--   'ConduitM'
+--
+-- [@Functions starting with o@]: This comes from \"mono-traversable\". The o
+--   choice is a bit strange, but it makes sense when you realize that the
```
snoyberg: This only appears to functions which are monomorphic, like `omap`. Also worth clarifying.

conduit/src/Data/Conduit/Combinators.hs
```diff
@@ -7,21 +7,18 @@
 {-# LANGUAGE NoImplicitPrelude         #-}
 {-# LANGUAGE NoMonomorphismRestriction #-}
 {-# LANGUAGE BangPatterns #-}
--- | This module is meant as a replacement for Data.Conduit.List.
```
snoyberg: It's worth adding a mention to the `Conduit` module. Note that it imports the unqualified names (with the `C` suffix).

PR #386, opened 2018-08-19
Implement 'MonadCatch' for 'ConduitT' and 'Pipe'

vapourismo:
Consider this more a question than an actual pull request. 

I couldn't help but wonder why `ConduitT` doesn't provide a `MonadCatch` instance (ignoring `catchC` here).
Apart from it breaking the efficiency gain from the codensity style, what are the reasons?

I am sure there is a good reason that I simply haven't discovered in my limited experience with `conduit`.

### DIFF
--- a/conduit/src/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/src/Data/Conduit/Internal/Conduit.hs
@@ -92,6 +92,7 @@ import Control.Exception (Exception)
 import qualified Control.Exception as E (catch)
 import Control.Monad (liftM, liftM2, ap)
 import Control.Monad.Error.Class(MonadError(..))
+import Control.Monad.Catch (MonadCatch (..), MonadThrow (..), try)
 import Control.Monad.Reader.Class(MonadReader(..))
 import Control.Monad.RWS.Class(MonadRWS())
 import Control.Monad.Writer.Class(MonadWriter(..), censor)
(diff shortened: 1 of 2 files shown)


snoyberg:
See discussion at https://github.com/yesodweb/yesod/issues/1533. tl;dr: I'm on the fence about adding these instances because of encouraging bad monad transformer practices.

Chobbes:
It seems like at least some conduit libraries rely upon `MonadThrow`, for e.g.:

https://hackage.haskell.org/package/xml-conduit-1.8.0.1/docs/Text-XML-Stream-Parse.html

And as there's no `MonadCatch` instance for `ConduitT`. This means there's no way to handle exceptions thrown by this library, right? And since the constructor for `ConduitT` is hidden I don't think you can add an orphan instance of `MonadCatch` for `ConduitT`.

I could be misunderstanding the problem, and it seems like exceptions are a bit of minefield where things might not behave as expected... But it seems like this would be useful for this kind of situation?

Is there a better solution to my problem? Like if I am using the XML library and I want to do something special when a parser raises an exception, what can I do?

snoyberg:
It depends entirely on the underlying monad. If you have `IO` underneath your `ConduitT`, then the exception handling functions from any runtime-exceptions library would work. You can also use conduit-specific exception handling functions like https://www.stackage.org/haddock/lts-14.17/conduit-1.3.1.1/Data-Conduit.html#v:tryC.

PR #403, opened 2019-01-24
Typo in README for operator using ZipConduit

bbarker:


snoyberg:
The code as written works for me, can you clarify why you believe this change is needed?

bbarker:
I admit I hadn't tried running the code so I must be wrong; the only other example in the codebase using ZipConduit I saw had the operator I changed it to `<*`. Hoogle reports a syntax error for `*>` but not `<*`. But maybe that's a Hoogle bug.

PR #406, opened 2019-02-18
Switch `gzip` to use zlib's special value of `-1` for compression level

lehins:
By using special value `-1` we let `zlib` know to use default compression level. It is denoted by `Z_DEFAULT_COMPRESSION = -1` in zlib and is currently equivalent to level `6`:

> Z_DEFAULT_COMPRESSION requests a default compromise between speed and compression (currently equivalent to level 6).


snoyberg:
Can you add a ChangeLog entry and minor version bump?

lehins:
@snoyberg Done.

PR #413, opened 2019-06-09
Add mapAsync to map a batch of actions asynchronously

bneijt:
Having used Akka streams in Scala, I was expecting something like `mapAsync`:
https://doc.akka.io/docs/akka/current/stream/operators/index.html#asynchronous-operators

But I could not get http://hackage.haskell.org/package/conduit-concurrent-map-0.1.1/docs/Data-Conduit-ConcurrentMap.html and could not find any such method in conduit-extra.

The solution below is trivial, but I would like to see methods like this in conduit-extra.

Is there interest to work towards adding utility methods like these to conduit-extra? If so, I would love to hear comments to move forward on this.

### DIFF
--- a/conduit-extra/Data/Conduit/Async.hs
+++ b/conduit-extra/Data/Conduit/Async.hs
@@ -1,0 +1,16 @@
+-- |
+-- Simple conduits to do async operations
+module Data.Conduit.Binary
+    (
+      mapAsyncIO
+    ) where
+
+import Conduit
+import qualified Data.Conduit.List as CL
+import Control.Concurrent.Async (mapConcurrently)
+
+-- | Map `a` asynchronously to `b` using batches of at most `n` elements using `mapConcurrently`
+mapAsync :: Int -> (a -> IO b) -> ConduitT a b IO ()
+mapAsync n op = CL.chunksOf n .| mapAsynced .| CL.concat
+    where
+        mapAsynced = mapMC (mapConcurrently op)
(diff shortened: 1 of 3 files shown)


snoyberg:
I think a pool of workers would make more sense here. As it stands, this will have strange behavior when, for example, one of the items in a chunk takes longer than others.

bneijt:
I have failed at the second attempt also. I thought maybe I could use a list of no op steps in my conduit to form a queue of un-awaited async actions, but that does not work.

I'm beginning to see why conduit-concurrent-map has such complicated worker/queue code.

I am going to give the problem a few more days and see if I can come up with something nice looking.

bneijt:
Well, I got it to work without a pool of workers and instead introducing a delayed conduit (yield only after another await). Affectively the conduit is now `spawn -> delay -> delay -> join` and because the delay step will pull another from the spawn, we end up with multiple threads running till they get to the join step.

This works and gives me multiple threads running at the same time.

The code could do with some clean-up and as I'm a beginner at Haskell you probably have a ton of operators to throw in there. Would this approach be something I could clean up enough to get it merged at some point in time?

snoyberg:
I haven't thoroughly read the code, but from what I've seen and your description, I'm overall -1 on this. Conduit is a half-duplex system, and trying to insert these kinds of things which require full-duplex functionality have always failed. I much prefer the approach of having a modified fusion function that allows full duplex.

bneijt:
Fair enough, thank you for looking into it.

PR #427, opened 2019-12-13
chunksOf: eagerly emit groups

chrismwendt:
Prior to this change, `chunksOf` did not emit groups as soon as it could:

```haskell
main = runConduit
  $ (yieldMany ['1', '2'] >> liftIO (print "should not print"))
  .| CL.chunksOf 2
  .| takeC 1
  .| printC
```

Prints

```
"should not print"
"12"
```

After this change, `chunksOf` will emit groups immediately when they are ready. Here's the new output:

```
"12"
```

Any tips on writing/updating tests?

Improves upon https://github.com/snoyberg/conduit/pull/296

### DIFF
--- a/conduit/ChangeLog.md
+++ b/conduit/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for conduit
 
+## 1.3.1.2
+
+* More eagerly emit groups in `chunksOf` [#427](https://github.com/snoyberg/conduit/pull/427)
+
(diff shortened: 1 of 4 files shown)


chrismwendt:
> Can you add a minor version bump and changelog entry?

Will do.

chrismwendt:
> Can you add a minor version bump and changelog entry?

Done. Also updated the tests, could you take a look?

PR #430, opened 2020-01-18
Add Acquire based version of bracketP.

merijn:
I'm working with an API that only exposes Acquire directly as means of working with a resource, the current bracketP in conduit makes it kinda nasty to work with. This PR adds an alternate version of bracketP working with Acquire values.

### DIFF
--- a/conduit/src/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/src/Data/Conduit/Internal/Conduit.hs
@@ -62,8 +62,8 @@ module Data.Conduit.Internal.Conduit
     , toProducer
     , toConsumer
       -- ** Cleanup
-    , bracketAcquireP
     , bracketP
+    , bracketAcquireP
       -- ** Exceptions
     , catchC
     , handleC
@@ -101,6 +101,7 @@ import Control.Monad.State.Class(MonadState(..))
 import Control.Monad.Trans.Class (MonadTrans (lift))
 import Control.Monad.IO.Unlift (MonadIO (liftIO), MonadUnliftIO, withRunInIO)
 import Control.Monad.Primitive (PrimMonad, PrimState, primitive)
+import Data.Acquire (Acquire, allocateAcquire, mkAcquire)
 import Data.Functor.Identity (Identity, runIdentity)
 import Data.Void (Void, absurd)
 import Data.Monoid (Monoid (mappend, mempty))
--- a/conduit/src/Data/Conduit/Internal/Conduit.hs
+++ b/conduit/src/Data/Conduit/Internal/Conduit.hs
@@ -62,6 +62,7 @@ module Data.Conduit.Internal.Conduit
     , toProducer
     , toConsumer
       -- ** Cleanup
+    , bracketAcquireP
     , bracketP
       -- ** Exceptions
     , catchC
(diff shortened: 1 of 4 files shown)


conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -896,11 +912,7 @@ bracketP :: MonadResource m
             -- ^ computation to run in-between
          -> ConduitT i o m r
             -- returns the value from the in-between computation
-bracketP alloc free inside = ConduitT $ \rest -> do
```
snoyberg: I'd rather leave this function unchanged.

conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -877,6 +879,20 @@ runConduit :: Monad m => ConduitT () Void m r -> m r
...
 
+-- | Version of 'bracketP' for working with libraries that only expose an
+-- 'Acauire' based API.
+--
+-- @since 1.3.2
+bracketAcquireP :: MonadResource m
+                => Acquire a
+                -> (a -> ConduitT i o m r)
+                -> ConduitT i o m r
+bracketAcquireP acquire inside = ConduitT $ \rest -> do
```
snoyberg: What's the reason for needing the underlying `ConduitT` representation here? Is it just for performance? I ask because I'm curious about what made `Acquire` difficult to use directly with the previous API.

merijn:
I used the conduit internal representation to copy bracketP, but upon further testing I think I misunderstood bracketP's claims of prompt finalization. The problem I was running into with Acquire was the finalizer not running when upstream wasn't fully consumed (i.e. when only consuming a small number of elements/peeking).

I understood bracketP's documentation to imply that it will run finalizers as soon as no further values will be requested from upstream, but it actually doesn't, which renders this change moot, as Acquire already handles the exception case and I will have to invent another solution for handling cases where a stream is only partially consumed.

snoyberg:
Conduit no longer has prompt finalization, explanation in this blog post: https://www.snoyman.com/blog/2018/01/drop-conduits-finalizers. If you see documentation that still talks about prompt finalization, a PR to improve it would be welcome.

Sounds like this PR can be closed though, do you agree?

merijn:
Yeah. Although this does raise some questions for me on how to correctly free resources on time. All the examples of `ResourceT` are fairly simplistic and I find it very hard to find documentation on how to correctly handle complex situations.

The problem I'm running into now is correctly freeing persistent queries in nested settings. To sketch a setting I have in my codebase:

```haskell
innerConduit :: (Monad m, ...) => ConduitT () Bar m Quux
innerConduit n = {- something like a query, consuming expensive resources -}

outerConduit :: (Monad m, MonadResource m) => ConduitT () Foo m r
outerConduit = do
    result <- innerConduit .| processBarsIntoFoo
    doStuffWith result
```

I'm trying to enforce that the resource allocated by `innerConduit` get released before `doStuffWith` starts running. If `processBarsIntoFoo` consumes all upstream elements this is fine, but if it only consumes a limited number of elements then any `release` calls inside `innerConduit` never run. Now, in non-conduit settings I'd just ensure that there's a ResourceT instance wrapped around `innerConduit .| processBarsIntoFoo` that I'd unwrap with `runResourceT` to bound the lifetimes of allocations, but that doesn't work here, as `runResourceT` requires `MonadUnliftIO` which `ConduitT` isn't an instance off.

The only way I can see to enforce prompt freeing of resources is by lifting the allocation out to `outerConduit` and calling `release` in between `innerConduit` and `doStuffWith`. But lifting every single resource allocation out of `innerConduit` is quite a hassle and kinda eliminates most of the composability I'm enjoying from `ConduitT`.

snoyberg:
I'd recommend turning `innerConduit` into a bracket-pattern function, something like:

```haskell
withInnerConduit :: (ConduitT i Bar m Quux -> ConduitT i o m a) -> ConduitT i o m a
withInnerConduit f = do
  (resource, key) <- allocate foo
  a <- f $ helper resource
  release key
  pure a
```

merijn:
Yeah, the downside of that approach is that if you use withInnerConduit with a downstream that doesn't consume all inputs you'd still leak the resource, so you always have to reason about the entire pipeline in it's entirety to be sure of finalisation, which I'd hoped to avoid, but I guess there's solution for that. (Well, maybe there's some way to extend ResourceT with some sort of "delimited region" so you can scope the allocation of resource somehow, but that becomes an entirely different PR/topic :)

PR #435, opened 2020-03-19
Monadic 'mapInput'

serras:
This PR adds a variant of `mapInput` which allows monadic effects to execute in the conversion. This turned out to be needed in [the Mu project](https://github.com/higherkindness/mu-haskell) to convert sinks where the conversion requires working in the `IO` monad.

### DIFF
--- a/conduit/conduit.cabal
+++ b/conduit/conduit.cabal
@@ -1,5 +1,5 @@
 Name:                conduit
-Version:             1.3.1.2
+Version:             1.3.1.3
 Synopsis:            Streaming data processing library.
 description:
     `conduit` is a solution to the streaming data problem, allowing for production,
--- a/conduit/conduit.cabal
+++ b/conduit/conduit.cabal
@@ -1,5 +1,5 @@
 Name:                conduit
-Version:             1.3.1.3
+Version:             1.3.2
 Synopsis:            Streaming data processing library.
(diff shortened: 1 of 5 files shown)


conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -989,6 +990,20 @@ mapInput f f' (ConduitT c0) = ConduitT $ \rest -> let
     go (Leftover p i) = maybe id (flip Leftover) (f' i) (go p)
     in go (c0 Done)
 
+-- | Apply a monadic action to all the input values of a @ConduitT@.
```
snoyberg: Please add a `@since` comment, ChangeLog entry, and minor version bump. Otherwise looks good, thanks!

serras: Done!

PS: This is the fastest review for a PR I've ever had 😸

conduit/conduit.cabal
```diff
@@ -1,5 +1,5 @@
 Name:                conduit
-Version:             1.3.1.2
+Version:             1.3.1.3
```
snoyberg: Following the PVP: minor version bump would bring this to 1.3.2, not 1.3.1.3.

Issue #85, opened 2013-02-10
"The mutable state is being accessed after cleanup" exception


nikita-volkov:
I have a concurrent program, which depending on some settings spits out: `Control.Monad.Trans.Resource.register': The mutable state is being accessed after cleanup. Please contact the maintainers.` 

I am highly unsure of where it's stemming from, and whether it's caused by my mistake or a bug of conduit or even by [this GHC bug](http://hackage.haskell.org/trac/ghc/ticket/7636) which the same program of mine suffers from in certain conditions.

Here is [a short screencast](http://screencast.com/t/jN10vAX7mF) describing the issue in detail. Could you please check it out?

MaxGabriel:
> Performing a forkIO inside of a ResourceT is inherently unsafe; you need to use resourceForkIO instead. It looks to me like there's a bug in couchdb's usage of resourcet, but without inspecting that codebase I can't be certain.

We've run into this a few times at work. I wanted to confirm this advice is still true. If so, should it be documented somewhere? I couldn't find anything about this in the haddocks, but I may have been looking in the wrong place. I could PR it

snoyberg:
I thought it was covered, but if not, or if you see places where the docs can be improved, a PR is definitely welcome.

PR #441, opened 2020-06-04
Document risk of using forkIO within a ResourceT

MaxGabriel:
Followup to conversation in #85

### DIFF
--- a/resourcet/Control/Monad/Trans/Resource.hs
+++ b/resourcet/Control/Monad/Trans/Resource.hs
@@ -265,9 +265,13 @@ resourceForkWith g (ResourceT f) =
 --
 -- This is defined as @resourceForkWith forkIO@.
 --
--- Note: Using regular 'forkIO' inside of a 'ResourceT' is inherently unsafe.
--- Using it is likely to lead to 'InvalidAccess' exceptions ("The mutable state is being accessed after cleanup").
--- Use this function, or other concurrency mechanisms like 'concurrently' or 'race', instead.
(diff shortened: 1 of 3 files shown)


MaxGabriel:
image

resourcet/Control/Monad/Trans/Resource.hs
```diff
@@ -265,6 +265,10 @@ resourceForkWith g (ResourceT f) =
 --
 -- This is defined as @resourceForkWith forkIO@.
 --
+-- Note: Using regular 'forkIO' inside of a 'ResourceT' is inherently unsafe.
```
MaxGabriel, line 268: Ideally this would end with ", since X" where X is a short, clear explanation of why it is unsafe, but I'm not familiar enough with resourceT to state that clearly

snoyberg: The issue is that, if you use `forkIO`, the child thread isn't registered as using the internal state. Then, when the parent thread stops using the internal state, it will automatically clean up all of the resources still in the state, even though the child thread may be using them.

When you use `resourceForkIO` instead, it increments a counter to let the parent thread know that someone else is using the state. Only when the last thread finally stops using the step will the counter go down to 0 and the resources be cleaned up, when we know that no one is using them.

MaxGabriel: Thanks, that's roughly what I thought, but wasn't positive. I updated the docs based on your explanation!

MaxGabriel: image

PR #443, opened 2020-06-24
Don't unmask resource allocation

andriusstank:
Asynchronous exceptions shouldn't be unmasked during resource allocation. Identical problem was fixed in `ResourceT` in 687df0555641489e8cf3e5d81cef9fd5d372be01.


snoyberg:
Thanks! Would you be able to add a minor version bump and changelog entry?

PR #446, opened 2020-08-31
Fix isChunksForExactlyE (fixes #445)

### LINKED ISSUE #445: Is chunksOfExactlyE correct in the case of an empty input chunk?
If I'm reading the definition of `chunksOfExactlyE` correctly, on the case `onull b` -- which means that the input chunk was empty -- `chunksOfE` will be called, changing the behavior to that of `chunksOfE` (with the remainder yielded and not left over). Am I reading this correctly? Is this the intended behavior for some reason, or is it a bug?

tilk:


snoyberg:
Thanks!

PR #447, opened 2020-09-08
Remove unneeded dependency on bytestring-builder

felixonmars:
It's not needed on newer GHC.


felixonmars:
Indeed. Updated.

snoyberg:
Thanks!

PR #451, opened 2020-10-15
Implement `uncons`

### LINKED ISSUE #442: Provide `uncons`
There is a more-or-less standard pattern:

```haskell
Y.unfoldM X.uncons
```

which can be used to convert a stream from a streaming library `X` to a streaming library `Y`. However, `conduit` doesn’t seem to provide `uncons` or any way of implementing it.

kirelagin:
Implement `uncons` (like for lists) and `unconsE` (returning result too) for pipes and conduits.

In the case of a conduit, we require it to be sealed, since what we essentially do is at each step take one element from the conduit and return the remainder, which the caller may later continue consuming one by one or not.

Resolves #442.


conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -720,6 +723,38 @@ connect :: Monad m
...
+--
+-- Note that you have to 'sealConduitT' it first.
+--
+-- Since 1.3.3
+uncons :: forall m o conduit. (Monad m, conduit ~ SealedConduitT () o m ())
+       => conduit -> m (Maybe (o, conduit))
+uncons (SealedConduitT p) = fmap (fmap (second SealedConduitT)) $ go p
+  where
+    -- This function is the same as @Pipe.uncons@ but it ignores leftovers.
+    go (HaveOutput p o) = pure $ Just (o, p)
```
snoyberg, line 736: If I'm reading this correctly, instead of the `fmap` and `second` business above, you could apply `SealedConduitT` to `p` on this line. Personally I'd find that easier to read. Is there a reason to avoid that?

conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -720,6 +723,38 @@ connect :: Monad m
         -> m r
 connect = ($$)
 
+-- | Split a conduit into head and tail.
+--
+-- Note that you have to 'sealConduitT' it first.
+--
+-- Since 1.3.3
+uncons :: forall m o conduit. (Monad m, conduit ~ SealedConduitT () o m ())
+       => conduit -> m (Maybe (o, conduit))
```
snoyberg, line 732: I don't think the equality constraint is a good thing here, it will make the docs harder to understand.

conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -720,6 +723,38 @@ connect :: Monad m
...
+    go (Done ()) = pure Nothing
+    go (PipeM mp) = mp >>= go
+    go (Leftover p ()) = go p
+
+-- | Split a coundit into head and tail or return its result if it is done.
+--
+-- Note that you have to 'sealConduitT' it first.
+--
+-- Since 1.3.3
+unconsE :: forall m o r conduit. (Monad m, conduit ~ SealedConduitT () o m r)
```
snoyberg, line 747: The distinction between `uncons` and `unconsE` doesn't seem to follow the standard naming scheme of "stream of chunks" vs "stream of values" found in the rest of the library

conduit/src/Data/Conduit/Internal/Pipe.hs
```diff
@@ -271,6 +273,32 @@ leftover = Leftover (Done ())
 {-# INLINE [1] leftover #-}
 {-# RULES "conduit: leftover l >> p" forall l (p :: Pipe l i o u m r). leftover l >> p = Leftover p l #-}
 
+-- | Split a pipe into head and tail.
+--
+-- Since 1.3.3
+uncons :: forall m o pipe. (Monad m, pipe ~ Pipe Void () o () m ())
```
snoyberg, line 279: Same comments as above apply here

snoyberg (changes requested):

conduit/src/Data/Conduit/Internal/Conduit.hs
kirelagin: Oops, I probably took this from some other streaming library...

kirelagin:
Fixed all of the above.

Also, as a side note, it would be nice to test that `unfoldM . unconsM === id` and `unfoldEitherM . unconsEitherM === id`, but I couldn’t find any utility that would compare two conduits by bisimulation and neither could I find an easy way to get both the outputs and the result from a conduit (which would be essentially bisimulation). Is there something like this?

snoyberg:
[`fuseBoth`](https://www.stackage.org/haddock/lts-16.18/conduit-1.3.2.1/Data-Conduit.html#v:fuseBoth) combined with `sinkList` will allow you to capture the outputs and result.

conduit/src/Data/Conduit/Internal/Conduit.hs
```diff
@@ -101,12 +103,13 @@ import Control.Monad.State.Class(MonadState(..))
 import Control.Monad.Trans.Class (MonadTrans (lift))
 import Control.Monad.IO.Unlift (MonadIO (liftIO), MonadUnliftIO, withRunInIO)
 import Control.Monad.Primitive (PrimMonad, PrimState, primitive)
+import Data.Bifunctor (second)
```
snoyberg, line 106: I think this is no longer necessary, correct?

snoyberg (changes requested):
Only one minor comment, looking good!

snoyberg (approved):
Looks great, thanks!

Issue #453, opened 2020-11-08
[RFC] foldWhile  


ocramz:
A specialized version of `mapAccumWhile` that keeps folding as long as the function argument returns 'Right'

```
foldWhile :: (Monad m) => (a -> s -> Either s s) -> s -> C.ConduitT a o m s
foldWhile f = loop
  where
    loop !s = C.await >>= maybe (pure s) go
      where
        go a = either (return $!) loop $ f a s
```

ocramz:
I found a need for this while writing a family of stochastic gradient descent algorithms; the conduit processes one data point of type `a` at a time and stops accumulating state when a convergence or divergence criterion is met.

snoyberg:
Seems reasonable to me. Only complaint I can think of: it hides away information about whether it was an end-of-stream versus a `Left` value produced by `f`. It may make more sense to generalize a bit to:

```haskell
foldWhile :: Monad m => (a -> s -> Either e s) -> s -> ConduitT a o m (Either e s)
```

Either with `fmap (either id id)` you recover the original behavior. Thoughts?

ocramz:
Indeed, your version of `foldWhile` is clearer. I'll prepare a PR!

PR #456, opened 2020-11-22
add foldWhile

ocramz:
Addresses #453

### DIFF
--- a/conduit/src/Data/Conduit/Combinators.hs
+++ b/conduit/src/Data/Conduit/Combinators.hs
@@ -1762,14 +1762,13 @@ STREAMING(mapAccumWhile, mapAccumWhileC, mapAccumWhileS, f s)
 -- Subject to fusion
 --
 -- @since 1.3.4
-foldWhile, foldWhileC :: Monad m => (a -> s -> Either e s) -> s -> ConduitT a o m (Either e s)
-foldWhileC f = loop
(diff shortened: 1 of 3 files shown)


conduit/src/Data/Conduit/Combinators.hs
```diff
@@ -1755,6 +1756,22 @@ mapAccumWhileC f =
 {-# INLINE mapAccumWhileC #-}
 STREAMING(mapAccumWhile, mapAccumWhileC, mapAccumWhileS, f s)
 
+
+-- | Specialized version of 'mapAccumWhile' that does not provide values downstream.
+--
+-- Subject to fusion
+--
+-- @since TODO
```
snoyberg, line 1764: Can you add a minor version bump, changelog entry, and fill in this `@since` comment?

conduit/src/Data/Conduit/Combinators.hs
```diff
@@ -1755,6 +1756,21 @@ mapAccumWhileC f =
 {-# INLINE mapAccumWhileC #-}
 STREAMING(mapAccumWhile, mapAccumWhileC, mapAccumWhileS, f s)
 
+
+-- | Specialized version of 'mapAccumWhile' that does not provide values downstream.
+--
+-- Subject to fusion
```
snoyberg, line 1762: I don't think this is true, since it isn't leveraging the STREAMING CPP. Can you remove it?

snoyberg (approved):
Thanks!

PR #462, opened 2021-03-28
Update README.md - Remove unnecessary import

fabfianda:
Remove unnecessary Product Monoid import in the example script.


snoyberg (approved):
Thanks!

PR #463, opened 2021-04-28
mkAcquire improvements

brandon-leapyear:


brandon-leapyear:
oh whoops guess it doesn't work after all. But maybe a helper function like `mkAcquireM`?

snoyberg:
This looks dangerous to me, it may encourage escaping the scope of the `MonadUnliftIO` monad. It we were to go in this direction, I'd recommend having an extra type `AcquireM m a` or similar, but I'm not convinced it's a good idea. Instead, I'd advocate for documenting how to use the existing API with `MonadUnliftIO`.

brandon-leapyear:
Makes sense to me. Just to be clear, escaping the scope is dangerous for monads like `ResourceT`, where the monad captures closable state, right? I would imagine most `MonadUnliftIO` monads use `ReaderT` without `IORef`s, in which case there isn't a problem with escaping scope.

snoyberg:
Yes, that's the case. But there are more of these "closable" things than just `ResourceT`. Even things like loggers may have cleanup that run at the end.

brandon-leapyear:
Sounds good. I added documentation for using `mkAcquire` with `MonadUnliftIO`.

snoyberg (approved):
Thanks!

PR #470, opened 2021-08-04
Fix forever leak

### LINKED ISSUE #469: `ResourceT` is subject to the `forever` space leak
```haskell
module Lib
    ( someFunc
    ) where

import Control.Monad

foreverM :: Monad m => m a -> m x
foreverM action = go
  where
    go = action >> go

someFunc :: IO ()
someFunc = runResourceT $ forever $ do
    pure ()

module Main where

import Lib
import Control.Concurrent
import System.Environment
import Control.Monad

main :: IO ()
main = do
    [n] <- fmap read <$> getArgs
    putStrLn "starting . . ."
    tid <- forkIO someFunc
    forM_ [1..n] $ \i -> do
        putStrLn $ show i <> " . . ."
        threadDelay (1 * 1000 * 1000)
    killThread tid
    putStrLn "done"
```

Compiling with `-O2`, and running with `forever` and `+RTS -s`, I'm seeing a linear increase in memory with the size of the loop.

Switching `forever` to `foreverM` fixes the problem, which means we're falling prey to the `forever` leak that plagued `transformers`.

parsonsmatt:
Closes #469 

I've verified that this fix does work with `forever` on the code in question.

### DIFF
--- a/resourcet/ChangeLog.md
+++ b/resourcet/ChangeLog.md
@@ -1,5 +1,10 @@
 # ChangeLog for resourcet
 
+
+## 1.2.4.3
+
+* Fix a space leak when using `forever` with `ResourceT`. [#470](https://github.com/snoyberg/conduit/pull/470)
+
 ## 1.2.4.2
 
 * Mask exceptions in `Acquire` allocation action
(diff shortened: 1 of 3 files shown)


snoyberg (approved):
Thanks!

Issue #404, opened 2019-01-27
Why were the MonadBase and MonadBaseControl instances removed from resourcet-1.2.0?


shlok:
In resourcet-1.1.11, we used to have the instances

    MonadBase b m => MonadBase b (ResourceT m)
    MonadBaseControl b m => MonadBaseControl b (ResourceT m)

Why were these removed in resourcet-1.2.0?

(I came across a situation where having the instances could have been useful: [`withMergedStreams` of Streaming.Concurrent](https://hackage.haskell.org/package/streaming-concurrency-0.3.1.2/docs/Streaming-Concurrent.html#g:3) requires these instances.)

snoyberg:
Please see https://www.snoyman.com/blog/2018/02/conduitpocalypse

shlok:
Thanks!

PR #477, opened 2022-02-20
Add MonadBase IO instances

chrismwendt:
This makes it easier to interface with [`lifted-async`](https://hackage.haskell.org/package/lifted-async), [`lifted-base`](https://hackage.haskell.org/package/lifted-base), and others.

The adds a dependency on `transformers-base`, which seems pretty small (107 LOC) and stable (~3 patch releases in 4 years).

### DIFF
--- a/conduit/conduit.cabal
+++ b/conduit/conduit.cabal
@@ -46,6 +46,7 @@ Library
   Build-depends:       base                     >= 4.12         && < 5
                      , resourcet                >= 1.2          && < 1.3
                      , transformers             >= 0.4
+                     , transformers-base
                      , mtl
                      , primitive
                      , unliftio-core
@@ -82,6 +83,7 @@ test-suite conduit-test
                    , hspec >= 1.3
                    , QuickCheck >= 2.7
                    , transformers
+                   , transformers-base
                    , mtl
                    , resourcet
                    , containers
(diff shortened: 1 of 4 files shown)


chrismwendt:
I stumbled upon https://github.com/snoyberg/conduit/issues/404#issuecomment-458418587 and discovered `unliftio` mostly does the trick 🎉 

However, there are so called "control" functions like `withMVar` which require `MonadUnliftIO`, but there's no instance for `MonadUnliftIO ConduitT i o m`, so apparently it's impossible to use `withMVar` inside of a `ConduitT`. Could this be implemented? What's preventing it from working? It might be hard to track the values being emitted from the callback's conduit, but it seems possible (although I haven't tried it).

As a fallback, could there be hybrid functions with plain `IO` callbacks like:

```haskell
withMVarHybrid :: MonadIO m => MVar a -> (a -> IO b) -> m b
withMVarHybrid v f = liftIO $ withMVar v f
```

Here's the unliftio implementation for comparison:

```haskell
withMVar :: MonadUnliftIO m => MVar a -> (a -> m b) -> m b
withMVar var f = withRunInIO $ \run -> M.withMVar var (run . f)
```

snoyberg:
You cannot set up "control" functions with `ConduitT` since it is continuation based. This is a deep topic, and I've written quite a bit on it previously. One starting point is:

https://github.com/snoyberg/conduit#resource-allocation

Adding `MonadBase` will not help with this, `MonadIO` already does the same thing and is insufficient for what you're trying to do.

chrismwendt:
Got it, I'll read up more before diving further into this.

PR #479, opened 2022-02-21
Add concurrentMap

chrismwendt:
Adds `concurrentMap`, which spawns worker threads to process values concurrently.

This adds dependencies on `containers` and `async`. If that's unacceptable, perhaps it could be moved to conduit-extra.

### DIFF
--- a/conduit/test/Spec.hs
+++ b/conduit/test/Spec.hs
@@ -44,7 +44,7 @@ import qualified Data.ByteString.Char8 as S8
 import qualified Data.ByteString.Lazy.Char8 as L8
 import qualified StreamSpec
 import GHC.IO (catchAny)
-import Control.Exception (fromException)
+import Control.Exception (fromException, Exception, SomeException, catch)
 import UnliftIO.Exception (pureTry, stringException)
 import Control.Concurrent.Async (ExceptionInLinkedThread (..))
 import Control.Concurrent.MVar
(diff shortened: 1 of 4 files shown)


conduit/test/Spec.hs
```diff
@@ -652,6 +655,44 @@ spec = do
...
+    describe "concurrentMap" $ do
+        let boom = stringException "boom"
+        let isBoom e = maybe False (== boom) $ fromException e
+        let isLinkedBoom e = case fromException e of
+                Nothing -> False
+                Just (ExceptionInLinkedThread _ inner) -> isBoom inner
+        let block = newEmptyMVar >>= takeMVar :: IO ()
+        it "propagates exceptions" $ do
+            let go = runConduitRes $ yield () .| concurrentMap 10 (\_ -> throwM boom) .| sinkNull
+            go `shouldThrow` isLinkedBoom
```
chrismwendt, line 667: This test is flaky and I haven't figured out why yet.

image

I verified `isLinkedBoom` returns `True` by wrapping it:

    go `catch` (\e -> print ("isLinkedBoom?", isLinkedBoom e) >> throwM e)

This one's got me stumped.

chrismwendt: I think it was caused by some race condition that would trigger 2 exceptions. The first exception would get caught, checked, and handled properly, but the second would strike outside the scope of the exception handlers (before or after).

I removed the `link worker` call (it's safe because all workers are `poll`ed, `wait`ed, or `cancel`ed), which means exceptions only bubble up when execution on the main thread is inside `concurrentMap` (e.g. if the main thread is in the source when a worker hits an exception then the main thread wouldn't know about it until the main thread yields a value from the source and `concurrentMap` reads it). I also tweaked the test to take that into account.

snoyberg:
So far we haven’t included this kind of behavior in conduit.  There’s a large area here for proper API design, and I’d be afraid of introducing potentially unstable APIs instead the core, stable library. I think it would be great to have a separate library focused on concurrency helpers for conduit.

chrismwendt:
Out of curiosity what might a proper API design look like? It seems like `concurrentMap`'s type signature is pretty good on its own, but maybe if there were a few more concurrency functions then a pattern of type signatures and naming conventions would emerge, and `concurrentMap` would be likely to change. Is that what you were thinking of too?

conduit-extra is "Experimental helper functions for conduit", so maybe that would be a good place for it. WDYT?

snoyberg:
It's not that there's any particular issue with the API design. It's that concurrent design usually ends up needing quite a bit of special tuning. For example: do we want to ensure that items stay in the same order? That right there can lead to an extra parameter or a parallel set of functions.

conduit-extra may say "experimental," but it isn't really that anymore, the API has been stable for years. Something that may include API changes should go in a new package so it can rapidly iterate.

chrismwendt:
Would a `Data.Conduit.Experimental` module be a possibility? It could be documented that this function is still undergoing fine tuning and might graduate to a non-experimental namespace once it's stable. As long as users aren't expecting it to remain stable, I don't see any downside.

It could even be documented in the Haddock docstring that the type signatures of anything in `*.Experimental` is subject to arbitrary change and doesn't follow semver.

I suppose I could make a separate package, but I'd worry a little about fragmenting the ecosystem further.

snoyberg:
No, I would rather it go into a separate package.

Issue #481, opened 2022-02-21
Implementing delay operator


emlautarom1:
I've been playing around with Conduit for the past few days and I'm looking for something like a 'delay'. For reference, I've had a pleasant experience using RxJS and one of the provided operators is [delay](https://rxjs.dev/api/operators/delay). I find it interesting that Conduit doesn't provide something like that out-of-the-box.

I implemented this kind of feature using the following:

```haskell
delay :: MonadIO m => Int -> ConduitT a a m ()
delay n = awaitForever $ \v -> do
  liftIO $ threadDelay n
  yield v
```

It looks very simple so I assume that something is wrong or I'm not considering some edge case. So far it has worked great, but I would like to know if this implementation is correct and if anyone is interested in adding it to the Combinators module.

snoyberg:
The implementation looks fine to me. This function doesn’t exist because no one’s ever expressed a need for it. I don’t see a problem with adding it if you’re interested in making a PR

emlautarom1:
Sure, I'll be glad to make a PR. I would appreciate any hint regarding tests for this operator!

snoyberg:
I’d recommend testing that at least the right amount of time has passed after rubbing a pipeline with delay.

PR #482, opened 2022-03-04
Add 'delay' combinator

emlautarom1:
Addresses https://github.com/snoyberg/conduit/issues/481


emlautarom1:
Don't really know why it fails on Windows...

Issue #392, opened 2018-09-13
About the UnliftIO.Resource module


mbwgh:
I am most likely confused about what the `UnliftIO.Resource` module is really about, but it is there that I would expect to be able to find more general versions of primitive functions like `allocate`.

For example, for my own purposes I wrote some

```haskell
allocate ::
  (MonadUnliftIO m, MonadResource m) => m a -> (a -> m ()) -> m (ReleaseKey, a)
```

since I wanted to be able to use `$logDebug` inside both the acquire and release parts, among other things. There is also the `Data.Acquire` module, but it too allows `MonadUnliftIO m => m a` actions to be executed inside the "inner" continuation only.

Have these kinds of generalizations intentionally been omitted, because they would break some laws? Or am I just using the library in a wrong way?

snoyberg:
To my knowledge, that specific generalization does not exist right now.

snoyberg:
I'd be OK with such as an addition, but it would need to go under a different name to avoid a conflict.

evanrelf:
I'm also interested in using this more generalized `allocate`. I believe it's as simple as:

```haskell
allocate'
  :: (MonadUnliftIO m, MonadResource m)
  => m a
  -> (a -> m ())
  -> m (ReleaseKey, a)
allocate' alloc free = withRunInIO $ \run ->
  run $ allocate (run alloc) (run . free)
```

Would you accept a PR for this? And if you want a different name, do you have any suggestions?

PR #490, opened 2022-06-29
Add unlifted version of `allocate`

evanrelf:
Closes [#392](https://github.com/snoyberg/conduit/issues/392).

If anyone has suggestions for a better name, let me know. I avoided
`allocate'` because the tick often implies some kind of strictness.


snoyberg:
Seems like a good name to me!

PR #491, opened 2022-07-04
Documentation: replace dead link with archived version

JaSpa:
The link in the documentation of `vectorBuilder` returns a 404:
https://www.fpcomplete.com/user/snoyberg/library-documentation/vectorbuilder. 

The Internet Archive though has a version available:
https://web.archive.org/web/20150919210626/https://www.fpcomplete.com/user/snoyberg/library-documentation/vectorbuilder


snoyberg (changes requested):
Good catch! But the content is still live, we just lost a redirect to its new home. I've sent a suggestion for the corrected URL.

snoyberg (approved):
Thanks!

PR #492, opened 2022-07-07
Remove reference to conduit-combinators

tonicebrian:
I was reading the documentation and searched conduit-combinators but there it says that the library is deprecated. This comment is misleading.


snoyberg (approved):
Agreed thanks!

PR #493, opened 2022-07-23
Update docs to avoid use of deprecated type aliases

avh4:
This inlines uses of the deprcated type aliases `Source`, `Sink`, `Producer`, `Consumer`, `Conduit` in the documentation and type annotations so that they now use the recommended `ConduitT` type instead.


snoyberg (approved):
This looks great, thank you!

PR #498, opened 2022-10-06
Include the exception in ReleaseTypes indicating exceptional exit.

### LINKED ISSUE #460: Acquire expose the ReleaseException exception

shlevy:
Supercedes and resolves #461.

Fixes #460.

Co-authored-by: Shea Levy <shea@shealevy.com>
Co-authored-by: parsonsmatt <parsonsmatt@gmail.com>

### DIFF
--- a/resourcet/Data/Acquire/Internal.hs
+++ b/resourcet/Data/Acquire/Internal.hs
@@ -26,13 +26,13 @@ import qualified Control.Monad.Catch as C ()
 -- @since 1.1.2
 data ReleaseType = ReleaseEarly
                  | ReleaseNormal
-                 | ReleaseException' E.SomeException
+                 | ReleaseExceptionWith E.SomeException
     deriving (Show, Typeable)
 
 {-# COMPLETE ReleaseEarly, ReleaseNormal, ReleaseException #-}
-{-# DEPRECATED ReleaseException "Use ReleaseException'" #-}
+{-# DEPRECATED ReleaseException "Use ReleaseExceptionWith" #-}
 pattern ReleaseException :: ReleaseType
-pattern ReleaseException <- ReleaseException' _
+pattern ReleaseException <- ReleaseExceptionWith _
 
 data Allocated a = Allocated !a !(ReleaseType -> IO ())
 
@@ -62,7 +62,7 @@ instance Monad Acquire where
     Acquire f >>= g' = Acquire $ \restore -> do
         Allocated x free1 <- f restore
         let Acquire g = g' x
-        Allocated y free2 <- g restore `E.catch` (\e -> free1 (ReleaseException' e) >> E.throwIO e)
+        Allocated y free2 <- g restore `E.catch` (\e -> free1 (ReleaseExceptionWith e) >> E.throwIO e)
         return $! Allocated y (\rt -> free2 rt `E.finally` free1 rt)
 
 instance MonadIO Acquire where
@@ -121,6 +121,6 @@ with :: MonadUnliftIO m
      -> m b
 with (Acquire f) g = withRunInIO $ \run -> E.mask $ \restore -> do
     Allocated x free <- f restore
-    res <- restore (run (g x)) `E.catch` (\e -> free (ReleaseException' e) >> E.throwIO e)
+    res <- restore (run (g x)) `E.catch` (\e -> free (ReleaseExceptionWith e) >> E.throwIO e)
     free ReleaseNormal
     return res
--- a/resourcet/Data/Acquire/Internal.hs
+++ b/resourcet/Data/Acquire/Internal.hs
@@ -30,7 +30,7 @@ data ReleaseType = ReleaseEarly
     deriving (Show, Typeable)
 
 {-# COMPLETE ReleaseEarly, ReleaseNormal, ReleaseException #-}
-{-# DEPRECATED ReleaseException "Use ReleaseExceptionWith" #-}
+{-# DEPRECATED ReleaseException "Use `ReleaseExceptionWith`, which has the exception in the constructor. This pattern synonym hides the exception and can obscure problems." #-}
 pattern ReleaseException :: ReleaseType
 pattern ReleaseException <- ReleaseExceptionWith _
 
--- a/resourcet/Data/Acquire/Internal.hs
+++ b/resourcet/Data/Acquire/Internal.hs
@@ -18,7 +18,7 @@ module Data.Acquire.Internal
 import Control.Applicative (Applicative (..))
 import Control.Monad.IO.Unlift (MonadIO (..), MonadUnliftIO, withRunInIO)
 import qualified Control.Exception as E
-import Data.Typeable (Typeable, typeOf)
+import Data.Typeable (Typeable)
 import Control.Monad (liftM, ap)
 import qualified Control.Monad.Catch as C ()
 
@@ -29,19 +29,6 @@ data ReleaseType = ReleaseEarly
                  | ReleaseNormal
                  | ReleaseException' E.SomeException
     deriving (Show, Typeable)
-
--- | Treats 'E.SomeException's as equal when they wrap the same type and 'show' the same.
-instance Eq ReleaseType where
-    ReleaseEarly == ReleaseEarly = True
-    ReleaseNormal == ReleaseNormal = True
-    ReleaseException' (E.SomeException e0) == ReleaseException' (E.SomeException e1) =
-        case typeOf e0 == typeOf e1 of
-            True ->
-                show e0 == show e1
-            False ->
-                False
-    _ == _ =
-        False
 
 -- | Fake 'E.Exception' to use with the deprecated 'ReleaseException' pattern.
 data DeprecatedReleaseExceptionPlaceholder = DeprecatedReleaseExceptionPlaceholder
--- a/resourcet/Data/Acquire/Internal.hs
+++ b/resourcet/Data/Acquire/Internal.hs
@@ -4,19 +4,21 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE PatternSynonyms #-}
 module Data.Acquire.Internal
     ( Acquire (..)
     , Allocated (..)
     , with
     , mkAcquire
-    , ReleaseType (..)
+    , ReleaseType (.., ReleaseException)
     , mkAcquireType
+    , DeprecatedReleaseExceptionPlaceholder
     ) where
 
 import Control.Applicative (Applicative (..))
 import Control.Monad.IO.Unlift (MonadIO (..), MonadUnliftIO, withRunInIO)
 import qualified Control.Exception as E
-import Data.Typeable (Typeable)
+import Data.Typeable (Typeable, typeOf)
 import Control.Monad (liftM, ap)
 import qualified Control.Monad.Catch as C ()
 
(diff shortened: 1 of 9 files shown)


resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,34 @@ import qualified Control.Monad.Catch as C ()
 -- @since 1.1.2
 data ReleaseType = ReleaseEarly
                  | ReleaseNormal
-                 | ReleaseException
-    deriving (Show, Read, Eq, Ord, Enum, Bounded, Typeable)
+                 | ReleaseException' E.SomeException
+    deriving (Show, Typeable)
+
+-- | Treats 'E.SomeException's as equal when they wrap the same type and 'show' the same.
+instance Eq ReleaseType where
```
snoyberg, line 34: I think I'm more tempted to simply bypass the instance entirely rather than rely on a potentially finicky `Show` instance, thoughts?

resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,34 @@ import qualified Control.Monad.Catch as C ()
...
+
+-- | Fake 'E.Exception' to use with the deprecated 'ReleaseException' pattern.
+data DeprecatedReleaseExceptionPlaceholder = DeprecatedReleaseExceptionPlaceholder
+     deriving (Show)
+
+instance E.Exception DeprecatedReleaseExceptionPlaceholder
+
+{-# COMPLETE ReleaseEarly, ReleaseNormal, ReleaseException #-}
+{-# DEPRECATED ReleaseException "Use ReleaseException'" #-}
+pattern ReleaseException :: ReleaseType
```
snoyberg, line 54: @parsonsmatt would you mind reviewing this part in particular?

snoyberg (commented):
Seems reasonable enough, I'd love a review by someone who understands pattern synonyms. Would you mind including a version bump and a changelog entry? I assume this needs to be a major version bump, right?

shlevy:
@snoyberg With the way the pattern synonyms are set up the major bump is only needed if you think the loss of instances on `ReleaseType` is important. Preexisting pattern matches and constructions will still be semantically correct.

resourcet/Data/Acquire/Internal.hs
shlevy: Done

resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,21 @@ import qualified Control.Monad.Catch as C ()
 -- @since 1.1.2
 data ReleaseType = ReleaseEarly
                  | ReleaseNormal
-                 | ReleaseException
-    deriving (Show, Read, Eq, Ord, Enum, Bounded, Typeable)
+                 | ReleaseException' E.SomeException
```
parsonsmatt, line 30: Gonna take issue with this name - the `'` feels unnecessary when a word can be more descriptive, and this isn't, like, "a strict `ReleaseException`", so the most common applicable mnemonic doesn't hold.

resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,21 @@ import qualified Control.Monad.Catch as C ()
...
+    deriving (Show, Typeable)
+
+-- | Fake 'E.Exception' to use with the deprecated 'ReleaseException' pattern.
+data DeprecatedReleaseExceptionPlaceholder = DeprecatedReleaseExceptionPlaceholder
+     deriving (Show)
+
+instance E.Exception DeprecatedReleaseExceptionPlaceholder
+
+{-# COMPLETE ReleaseEarly, ReleaseNormal, ReleaseException #-}
```
parsonsmatt, line 39: This `COMPLETE` pragma means that GHC will not issue a warning when you do the following:

```haskell
case x of
    ReleaseEarly -> print 1
    ReleaseNormal -> print 2
    ReleaseException -> print 3
```

Without the pragma, GHC has no way of knowing that `ReleaseException` handles the `ReleaseException'` case, and will complain that you aren't handling `ReleaseException' _`.

GHC does not check this pragma - it trusts it.

resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,21 @@ import qualified Control.Monad.Catch as C ()
...
+{-# DEPRECATED ReleaseException "Use ReleaseException'" #-}
+pattern ReleaseException :: ReleaseType
+pattern ReleaseException <- ReleaseException' _
+  where
+    ReleaseException = ReleaseException' (E.toException DeprecatedReleaseExceptionPlaceholder)
```
parsonsmatt, line 44: This is a bidirectional pattern synonym, meaning it can be used both for *constructing* a value and *pattern matching* on it.

The first line:

```haskell
pattern ReleaseException <- ReleaseException' _
```

describes how the pattern match works. When you match on `ReleaseException`, it will 'replace' that with a match on `ReleaseException' _` - this catch-all just ignores the new parameter.

parsonsmatt: The second part,

```haskell
  where
    ReleaseException = ReleaseException' (E.toException DeprecatedReleaseExceptionPlaceholder)
```

describes what happens when you use this to *construct* with the term. Let's consider this definition:

```haskell
foo :: ReleaseType
foo = ReleaseException
```

Before this code change, we'd have a simple value - `ReleaseException :: ReleaseType`. After this change, the syntax doesn't change, but the actual value that gets constructed becomes `ReleaseException' (SomeException DeprecatedReleaseExceptionPlaceholder)`.

---

I'm unsure how I feel about having the pattern synonym for construction as well as matching. Constructing a `ReleaseType` should really rarely happen, and as a library user, I'd *much* rather have an error pointing me towards `ReleaseExceptionWith`, so I can update my code and provide the exception.

Right now, that notice is given using the `DEPRECATION` pragma on `ReleaseException`, pointing us towards the new `ReleaseException'`. But this means that you'll get warnings (and `-Werror` build failures) when pattern matching on `ReleaseException`, which is most likely much more common for end-users to do than constructing.

If we drop the construction half of the synonym, then we don't need to deprecate the pattern synonym, and I'd guess almost all end users will have no new warnings or errors. The downside of this is that end users won't know about this change unless they go looking for it.

Briefly, the two choices appear to be:

- All users that reference the constructors get `DEPRECATION` warnings
- Users that *construct* a `ReleaseException` get a compile-time error, and users that match on a `ReleaseException` continue on as-usual.

resourcet/test/main.hs
```diff
@@ -149,3 +149,26 @@ instance Exception Dummy
...
+
+releaseNormal :: ReleaseType -> Bool
+releaseNormal ReleaseNormal = True
+releaseNormal _ = False
+
+releaseException :: (Exception e) => Selector e -> ReleaseType -> Bool
+releaseException sel (ReleaseException' se) = case fromException se of
+                         Just e -> sel e
+                         Nothing -> False
+releaseException _ _ = False
```
parsonsmatt, line 167: ah, yeah, that's why i had the `Eq` instance, lol

resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,21 @@ import qualified Control.Monad.Catch as C ()
...
-                 | ReleaseException
-    deriving (Show, Read, Eq, Ord, Enum, Bounded, Typeable)
+                 | ReleaseException' E.SomeException
+    deriving (Show, Typeable)
+
+-- | Fake 'E.Exception' to use with the deprecated 'ReleaseException' pattern.
+data DeprecatedReleaseExceptionPlaceholder = DeprecatedReleaseExceptionPlaceholder
+     deriving (Show)
+
+instance E.Exception DeprecatedReleaseExceptionPlaceholder
```
parsonsmatt, line 37: If we keep construction, then we should probably attach a `DEPRECATION` pragma to this type too

resourcet/Data/Acquire/Internal.hs
shlevy: No strong opinions.

resourcet/Data/Acquire/Internal.hs
shlevy: There's also:

* Users that *construct* a `ReleaseException` get a compile-time error, and users that match on a `ReleaseException` get a `DEPRECATION` warning

I see the argument for making the pattern uni-directional, but it's not clear why we still wouldn't want users to move to the proper constructor in matching.

resourcet/Data/Acquire/Internal.hs
shlevy: Why? The only reason a user would want to reference this is if they wanted to know that they're downstream of broken code, but in that case this isn't deprecated, there's no replacement.

parsonsmatt: Suppose we complete the deprecation by deleting the `ReleaseException` pattern. Then there is no reason to have this around either, and it gets deleted. I'd call that 'deprecated'.

resourcet/Data/Acquire/Internal.hs
parsonsmatt: I guess there are two more options:

- Users don't get any errors or warnings (no `DEPRECATION` message + bidirectional synonym)
- Users get compile-time errors for pattern matching or construction (no `PatternSynonym` at all)

From a "no breaking changes" perspective, "no errors or warnings" satisfies it pretty well - lacking the deprecation warning, this variant isn't even a major breaking change.

But it has a cost - the end-user will not be aware of two things:

1. If the end user is constructing `ReleaseException`s, then they will continue to (unknowingly) construct dummy values that drop useful information.
2. If the end user is *matching* on `ReleaseException`, and they want to know about the exception, then they won't know about it.

I'm more worried about the downstream effects of 1. Having to worry about dummy exceptions is really annoying, all so some weird intermediate library scenario can continue to sort-of have backwards compatibility.

On the other hand, "no pattern synonym at all" means that everyone gets a compile error. This means that folks will be made aware of these two bits of information as they fix their build. BUT, they have to fix their code. And then you have the whole major version hullabaloo with CPP to support multiple versions.

---

So, here's a middle way:

1. Introduce the bidirectional pattern synonym, *without a deprecation*, as a *minor* version bump. Since we're not introducing new warnings or errors, and pattern matches will all be `COMPLETE` still, I think this is fine. This version bump can be a 'compatibility' version.
2. Immediately afterwards, deprecate the `ReleaseException` pattern synonym with an upgrade path message. This is a major version bump.
3. At the *next* major version, delete the synonym and other deprecated terms.

With this, end users can depend on `resourcet-1.2.7` (next minor version). Then `resourcet-1.3.0` has the warnings, and `resourcet-1.4.0` removes the synonym entirely.

Libraries that require `resourcet <= 1.3` won't block end users from getting access to the exception, so we can deliver the feature without blocking anyone.

parsonsmatt: All of these "bad case" scenarios are pretty unlikely though. I think any solution with a pattern synonym here will prevent any major pain or churn in the ecosystem.

resourcet/Data/Acquire/Internal.hs
```diff
@@ -25,8 +27,21 @@ import qualified Control.Monad.Catch as C ()
...
+
+{-# COMPLETE ReleaseEarly, ReleaseNormal, ReleaseException #-}
+{-# DEPRECATED ReleaseException "Use ReleaseException'" #-}
```
parsonsmatt, line 40:
```suggestion
{-# DEPRECATED ReleaseException "Use `ReleaseException'` which has the exception in the constructor. This pattern synonym hides the exception and can obscure problems." #-}
```

shlevy:
A bit of back and forth in the comments, the big open question is the transition path to the new constructor. There are a bunch of possible permutations, but here are what I think the major candidates are, in my preference order:

1. Simply add the field to `ReleaseException`, bump the major version, and expect the likely few cases that match on `ReleaseException` and the quite possibly zero cases that construct `ReleaseException` to just deal with the compilation failures. This makes for a clean interface, avoids complication down the road, and the impact is likely fairly minimal.
2. Like 1 but retain the pattern synonym for *matching* on `ReleaseException` but not one for constructing. I wouldn't be surprised if there are literally 0 real-world projects that construct `ReleaseException` besides `resourcet` itself. I think it would make sense to deprecate the pattern synonym in this case, and perhaps remove it later. I think this is basically what @parsonsmatt recommends, though I'm not sure he likes the deprecation.
3. Have both pattern synonyms, together with the fake `Exception`. I don't really think this would even require a major version bump, but it would mean that code which relies on the exception would have to check for the dummy and it's probably a theoretical issue. This is what is currently implemented.

resourcet/Data/Acquire/Internal.hs
shlevy: > lacking the deprecation warning, this variant isn't even a major breaking change.

I don't think a deprecation should move a minor bump to a major one.

shlevy: > And then you have the whole major version hullabaloo with CPP to support multiple versions.

Nah, just wildcard match if you want to support both types and don't care about the exception.

snoyberg:
@parsonsmatt do you have a preference here? I feel pretty far out of depth here given my lack of experience with pattern synonyms. I'd probably in a vacuum lean towards (1) just due to my inexperience.

parsonsmatt:
I think we'll have the least ecosystem pain with scenario 2. That's my recommendation. 

Deprecating the pattern synonym is fine - libraries don't build with `-Werror` and applications probably want the feature anyway.

shlevy:
Updated to match @parsonsmatt's recommendation.

resourcet/Data/Acquire/Internal.hs
shlevy: Done.

resourcet/Data/Acquire/Internal.hs
shlevy: Fake exception removed.

snoyberg (approved):
This all looks good to me, I'm happy to merge/release at this point. @parsonsmatt any final comments before moving ahead with it?

Thanks to both you and @shlevy for hashing the details out so thoroughly here.

shlevy:
@snoyberg Good to go?

shlevy:
Also, can we get a release once this is in? Would like to depend on it in the next release of my lib.

snoyberg:
Release to Hackage uploaded.

Bodigrim:
As a Hackage trustee I made a matching revision of `cereal-conduit` to allow `resourcet-1.3`: https://hackage.haskell.org/package/cereal-conduit-0.8.0/revisions/

PR #501, opened 2023-02-16
add missing MonadCatch instance

ners:
`ConduitT` implements `MonadError` for a generic error type, which is in principle equivalent to `MonadCatch`.

Tested this together with @vkleen on https://github.com/ners/dosh


snoyberg:
Please see prior discussions, it's by design right now that we don't have a MonadCatch instance: https://github.com/snoyberg/conduit/pull/386.

PR #513, opened 2024-08-26
Avoid Dropping Upstream Items in `mergeSource`

julmb:
From what I understand, `mergeSource` will `await` an item from upstream, and if successful, `await` an item from `src0`. If there is no item in `src0`, the conduit terminates. This means that the item that was already taken from upstream is now lost (note the missing item `4` in the output):
```hs
>>> let num = C.yieldMany [1 .. 10]
>>> let chr = C.yieldMany ['a' .. 'c']
>>> runConduitPure $ num .| liftA2 (,) (mergeSource chr .| C.sinkList) C.sinkList
([('a',1),('b',2),('c',3)],[5,6,7,8,9,10])
```

This PR adds the already consumed item back using `leftover`, leading to the expected result:
```hs
([('a',1),('b',2),('c',3)],[4,5,6,7,8,9,10])
```

Disclaimer: I am pretty new to Conduit, so it is possible that I am either misunderstanding something or that there is a good reason for the current behavior of `mergeSource`, in which case, feel free to correct me.

### DIFF
--- a/conduit/ChangeLog.md
+++ b/conduit/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for conduit
 
+## 1.3.6
+
+* Avoid dropping upstream items in `mergeSource` [#513](https://github.com/snoyberg/conduit/pull/513)
+
 ## 1.3.5
 
 * Add `groupOn`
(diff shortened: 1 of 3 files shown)


snoyberg (commented):
Seems like a good change. Can you add a comment to the changelog and do a minor version bump? An added test case would be great too, but not a blocker.

snoyberg (approved):
Thanks!

snoyberg:
There's a build error in the tests due to missing the `liftA2` import, can you address that?

julmb:
I pushed right as you wrote that, let's hope it works now. ;)

snoyberg:
LGTM, thanks!

PR #519, opened 2025-02-23
Remove noncanonical definitions

rhendric:
This appeases the -Wnoncanonical-monad-instances warning. This warning exists because a future GHC release may treat noncanonical definitions as errors.

See also:
  https://gitlab.haskell.org/ghc/ghc/-/wikis/proposal/monad-of-no-return

(Special note: ordinarily, the remedy here would be to remove the `(*>) = (>>)` definition. But snoyberg/conduit#497 indicates that this would cause a performance regression, so instead I have inlined `(>>)`. The important thing is that, if `(>>)` is later defined to equal `(*>)`, this doesn't result in a loop.)

### DIFF
--- a/conduit/ChangeLog.md
+++ b/conduit/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for conduit
 
+## 1.3.6.1
+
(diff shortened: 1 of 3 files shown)


snoyberg (approved):
Thanks!

PR #524, opened 2025-06-05
Gracefully handle subprocess closing stdin

### LINKED ISSUE #523: sourceProcessWithStreams crashes if process closes input stream
Running `vips` on a `.bmp` file via Data.Conduit.Process.sourceProcessWithStreams
fails incorrect ResourceVanish IO exception.

I've created a [minimal reproduction](https://github.com/janvogt/conduit-vips-bmp-ResourceVanished) that exposes this interaction.

Running `./repro.sh` within the provided environment fails with 

> Repro.hs: sample1.bmp: withBinaryFile: resource vanished (Broken pipe)

despite the file clearly existing.

Also it has to be the interaction, of these three aspects. Every aspect on its own works as shown as well in the repro:

- `sourceProcessWithStreams` works with `cat` on `.bmp` and all other files tested
- `sourceProcessWithStreams` works with `vips` on `.heic` and all files that are not `.bmp`
- `vips` on the command line works with `.bmp` and all other files tested

Only `.bmp` files processed with `vips` using `sourceProcessWithStreams` fails afaick.

janvogt:
This fixes #523

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@ tarballs/
 .stack-work/
 *~
 *#
+
+# No nix & direnv setup
+flake.nix
+flake.lock
+.envrc
+.direnv
--- a/conduit-extra/ChangeLog.md
+++ b/conduit-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for conduit-extra
 
+## 1.3.8
+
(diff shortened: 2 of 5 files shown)


snoyberg:
Thanks!

PR #526, opened 2025-06-30
Base 4.14

### LINKED ISSUE #525: Bounds for conduit-extra seem off
Hi there,

The newest version of conduit-extra doesn't seem build on GHC 8.8, even though it still has pragmas to support 7.10?

At issue is the use of `isResourceVanishedErrorType` which was only introduced in base 4.14. Maybe a hackage update on the bounds would be good?

Cheers,
Huw

JackKelly-Bellroy:
Closes #525

Can we please have a metadata revision for `base >= 4.14` against `conduit-extra-1.3.8`?

### DIFF
--- a/conduit-extra/ChangeLog.md
+++ b/conduit-extra/ChangeLog.md
@@ -1,5 +1,9 @@
 # ChangeLog for conduit-extra
 
+## Unreleased (patch)
+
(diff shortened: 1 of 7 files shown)


JackKelly-Bellroy:
The discussion for #525 notes that metadata has been revised, but it'd still be great if this PR could get reviewed and merged.

Issue #459, opened 2021-02-19
sourceDirectoryDeep False could be much faster on Linux for large directories


nh2:
Just writing this down somewhere, with a suggestion for the solution:

[`sourceDirectoryDeep`](https://github.com/snoyberg/conduit/blob/0bf72ae4c5a765517ef69df650456c11b1e8fbf1/conduit/src/Data/Conduit/Combinators.hs#L729) unnecessarily runs the `stat()` syscall on every file, thus making it slow on large directories or directory trees. For example, it takes 30 minutes on my server to run it on a directory tree with 8M files. `find` does it in 30 seconds. [`sourceDirectory`](https://github.com/snoyberg/conduit/blob/0bf72ae4c5a765517ef69df650456c11b1e8fbf1/conduit/src/Data/Conduit/Combinators.hs#L704) does not have this problem, but cannot traverse into subdirs. On Linux, the problem can be observed with `strace`.

If it didn't `stat()`, only using `getdents64()` to list the directory contents (which retrieves multiple file names in a batch), it would be at least 60 times faster.

The `stat()` happens in the call to `getFileType` here: 

https://github.com/snoyberg/conduit/blob/0bf72ae4c5a765517ef69df650456c11b1e8fbf1/conduit/src/Data/Conduit/Combinators.hs#L737

It does that to distinguish normal files from symlinks to implement the `followSymlinks :: Bool` feature.

However the `find` UNIX utility also doesn't `stat()` (unless the `-L` option is given to follow symlinks); yet it can traverse into subdirectories. How does it do it?

First the call graph:

* [`sourceDirectory` calls `openDirStream`+`readDirStream`](https://github.com/snoyberg/conduit/blob/0bf72ae4c5a765517ef69df650456c11b1e8fbf1/conduit/src/Data/Conduit/Combinators.hs#L711) from [`streaming-commons`](https://hackage.haskell.org/package/streaming-commons-0.2.2.1/docs/Data-Streaming-Filesystem.html)
  * which calls [`System.Posix.Directory.openDirStream`+`readDirStream`](https://hackage.haskell.org/package/unix-2.7.2.2/docs/System-Posix-Directory.html#v:openDirStream) from `unix`, which calls `c_opendir`/`c_readdir`
    * which are `foreign import`s of `__hscore_opendir`+`__hscore_readdir`
[...]

PR #527, opened 2025-07-06
Make use of dirEntType in sourceDirectoryDeep

sjshuck:
This is halfway towards #459.  The other half is getting `dirEntType` itself fixed in unix, which is currently hard-coded to return `UnknownType` and effectively reverts to the current behavior of `stat()`ing every file.  That fix is https://github.com/haskell/unix/pull/348.  They can be merged in any order.

Benchmarks running `sourceDirectoryDeep` into `sinkNull` from my home dir (about 500K files):

| | Follow symlinks? | Time |
| --- | --- | --- |
| Current | &cross; | 1.71s |
| Current | &check; | 3.19s |
| After this PR | &cross; | 1.70s |
| After this PR | &check; | 3.18s |
| After this and the unix PR | &cross; | 0.94s |
| After this and the unix PR | &check; | 1.96s |

### DIFF
--- a/conduit/benchmarks/source-directory.hs
+++ b/conduit/benchmarks/source-directory.hs
@@ -1,0 +1,16 @@
+import Conduit
+import System.Directory (getHomeDirectory)
+import Test.Tasty.Bench
+
+main :: IO ()
+main = defaultMain
+  [ env getHomeDirectory $ \home ->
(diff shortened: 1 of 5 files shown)


sjshuck:
Bump

